PR #107, opened 2015-01-03
devel_glfw3.1: Update to glfw/glfw@aa7f827e72.

dmitshur:
Pretty straightforward, I didn't have to fix anything to get it to compile.

Been using it locally the last few days with no issues.

I'm really looking forward to the `direct` branch being merged into `glfw/glfw`'s `master`, as that will fix many issues (see https://github.com/glfw/glfw/commit/3ab6126e937c33f484534be4ff3547c21f39a286). But that hasn't happened yet.

I would prefer to merge this PR myself (without a merge commit), just need a LGTM comment.


emidoots:
LGTM.

I guess for mir support we will also need a `mir` build tag like our current `wayland` one?

dmitshur:
I think so.

But we need to investigate/be careful about how that build tag propagates. See https://github.com/libgit2/git2go/pull/159#commitcomment-9124040 for an issue that `git2go` people discovered when they tried something similar. However, we're building statically as far as I can tell this may not affect us. Need to confirm.

PR #111, opened 2015-01-18
Replace error codes with more idiomatic Go error values.

emidoots:
This replaces the C-style GLFW error codes with more idiomatic Go error values.

Instead of writing an error check like:

```
if err.Code == glfw.NotInitialized {
}
```

you would just write:

```
if err == glfw.ErrNotInitialized {
}
```

### DIFF
--- a/error.go
+++ b/error.go
@@ -8,49 +8,103 @@ import (
 	"fmt"
 )
 
-// ErrorCode corresponds to an error code.
-type ErrorCode int
+// prefixedError is a error that returns a error string in the format of:
+//
+//  prefix: desc
+//
+type prefixedError struct {
+	prefix, desc string
+}
 
-// Error codes.
-const (
-	NotInitialized     ErrorCode = C.GLFW_NOT_INITIALIZED     // GLFW has not been initialized.
-	NoCurrentContext   ErrorCode = C.GLFW_NO_CURRENT_CONTEXT  // No context is current.
-	InvalidEnum        ErrorCode = C.GLFW_INVALID_ENUM        // One of the enum parameters for the function was given an invalid enum.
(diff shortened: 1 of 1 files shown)


dmitshur:
Okay, I think the API change here is good, and being able to do `if err == glfw.ErrSomething` instead of `if err.Code == glfw.Something` is an improvement.

There's a major problem right now in that it doesn't actually work. Consider:

``` Go
package main

import (
    "fmt"

    "github.com/go-gl/glfw3" // on 3.1_no_error_codes branch
)

func main() {
    _, err := glfw3.CreateWindow(400, 400, "", nil, nil)
    if err != nil {
        fmt.Println("err == glfw3.ErrNotInitialized:", err == glfw3.ErrNotInitialized)

        fmt.Println(err)
    }

    // Output:
    // err == glfw3.ErrNotInitialized: false
    // ErrNotInitialized: The GLFW library is not initialized
}
```

The problem is that `glfw.ErrNotInitialized` is a _value_, and `newError` returns a copy of that value, but modifies its `desc` field, so equality doesn't hold.

The idiomatic `err == library.ErrSomething` stuff works when `ErrSomething` is defined as an unchanging value, like `errors.New("...")`. If you're modifying `ErrSomething`, equality checks won't work.

emidoots:
The fact that they're values and unequal is an aspect I didn't consider.

As you discussed with my on Slack, our only alternatives would be:
- Pointers w/modifying the value, but that'd make them unsafe in multi-threaded applications.
- Interfaces or multiple `os.IsNotExist(err error)`-esque functions, which sound clunky.
- Stick with what we have now.

I'm going to close this PR and say we stick with error codes as we have now.

PR #112, opened 2015-01-19
3.1 clean errors

emidoots:
This should be my last push for more clean/idiomatic errors from GLFW.

Please see the individual commit descriptions for more information.

### DIFF
--- a/error.go
+++ b/error.go
@@ -11,13 +11,18 @@ import (
 // ErrorCode corresponds to an error code.
 type ErrorCode int
 
+// Error codes that are translated to panics and the programmer should not
+// expect to handle.
+const (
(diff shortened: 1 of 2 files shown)


emidoots:
> The commit message doesn't make sense:

Sorry, my mistake. I've sent an alternative PR with the typo fixed.

error.go
```diff
@@ -75,7 +105,12 @@ func flushErrors() {
 func fetchError() error {
 	select {
 	case err := <-lastError:
-		return err
+		switch err.Code {
+		case outOfMemory, invalidEnum, invalidValue, notInitialized, noCurrentContext:
```
shurcooL: It seems a little odd these are not in the same order as above, but I guess that's fine.

PR #113, opened 2015-01-19
3.1 clean errors

emidoots:
This should be my last push for more clean/idiomatic errors from GLFW.

Please see the individual commit descriptions for more information. 

Alternative to #112 (fixed typo in commit message)

### DIFF
--- a/error.go
+++ b/error.go
@@ -11,13 +11,18 @@ import (
 // ErrorCode corresponds to an error code.
 type ErrorCode int
 
+// Error codes that are translated to panics and the programmer should not
+// expect to handle.
+const (
+	notInitialized   ErrorCode = C.GLFW_NOT_INITIALIZED    // GLFW has not been initialized.
+	noCurrentContext ErrorCode = C.GLFW_NO_CURRENT_CONTEXT // No context is current.
+	invalidEnum      ErrorCode = C.GLFW_INVALID_ENUM       // One of the enum parameters for the function was given an invalid enum.
+	invalidValue     ErrorCode = C.GLFW_INVALID_VALUE      // One of the parameters for the function was given an invalid value.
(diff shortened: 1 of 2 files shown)


error.go
```diff
@@ -75,7 +105,12 @@ func flushErrors() {
 func fetchError() error {
 	select {
 	case err := <-lastError:
-		return err
+		switch err.Code {
+		case outOfMemory, invalidEnum, invalidValue, notInitialized, noCurrentContext:
```
dmitshur: It seems a little odd these are not in the same order as above, but I guess that's fine.

dmitshur:
LGTM.

tapir:
:thumbsup: LGTM

tapir:
should we add the error anlaysis here (http://www.glfw.org/docs/latest/group__errors.html) to go doc documentation?

error.go
emidoots: Done.

emidoots:
According to the doc link @tapir sent, `GLFW_PLATFORM_ERROR` is:

> A bug in GLFW or the underlying operating system. Report the bug to our issue tracker.

On that premise, I'll also turned that one into a panic.

tapir:
The thing is now all the functions return an error and most of the errors are actually panics.
Looks a bit overdone. Ideas?

emidoots:
> should we add the error anlaysis here (http://www.glfw.org/docs/latest/group__errors.html) to go doc documentation?

Done.

PTAL again everyone.

emidoots:
> The thing is now all the functions return an error and most of the errors are actually panics.
> Looks a bit overdone. Ideas?

Let's not discuss this here: I'll create a new issue to discuss it.

tapir:
LGTM

dmitshur:
LGTM.

PR #114, opened 2015-01-19
Remove a typo that is kept only for backward compatibility

tapir:


dmitshur:
Good catch, LGTM.

emidoots:
nice

LGTM.

PR #116, opened 2015-01-19
Standard cursor

tapir:


tapir:
Any ideas about the name collision?

pwaller:
Your const was colliding with this type declaration:

https://github.com/go-gl/glfw3/commit/e69a1763102f13163ea68b3d1c392cd5828cec98#diff-3af9fbbb4c0bf6c982b7cac80cbbfaa4R245

If you want to be able to make sense of the errors, you can run `go build -v -x -work` and then inspect the file `/tmp/go-build446714385/github.com/go-gl/glfw3/_obj/input.cgo1.go:253` to look at the line numbers.

tapir:
I know I meant do you guys like the name change? I've added "Mode" at the end of consts. But I'm not sure if this is OK since it's not same as C API.

emidoots:
I am a bit tired, so I might be wrong. But I don't see a way to initialized the pixel data for a `Image` ?

Also if it's 32bpp little-endian RGBA, why aren't we using `*image.RGBA` directly instead?

tapir:
You're right I should implement the struct
properly. I will check image.RGBA also.

2015-01-19 9:46 GMT+01:00 Stephen Gutekanst notifications@github.com:

tapir:
Latest commit tried and working with this example

``` go
package main

import (
    "runtime"
    "image"
    _ "image/png"
    "os"
    glfw "glfw3"
)

func init() {
    runtime.LockOSThread()
}

func main() {
    err := glfw.Init()
    if err != nil {
        panic(err)
    }
    defer glfw.Terminate()

    window, err := glfw.CreateWindow(640, 480, "Testing", nil, nil)
    if err != nil {
        panic(err)
    }
    window.MakeContextCurrent()

    reader, err := os.Open("cursor.png")
    if err != nil {
       panic(err)
    }
    defer reader.Close()

    m, _, err := image.Decode(reader)
    if err != nil {
        panic(err)
    }

    c, err := glfw.CreateCursor(glfw.Image(m.(*image.NRGBA)), 0, 0)
    if err != nil {
        panic(err)
    }

    err = window.SetCursor(c)
    if err != nil {
        panic(err)
    }

    for isFinished := false; !isFinished; isFinished, _ = window.ShouldClose() {
        // Do OpenGL stuff
        window.SwapBuffers()
        glfw.PollEvents()
    }
}
```

Need more clarification about NRGBA vs RGBA. From my experiments if I create an image with transparency, Go recognizes it as NRGBA, if it doesn't have any transparent pixels, it recognizes it as RGBA. I'm not an expert on this. Opinions are welcome.

tapir:
What we can do though do a type switch and convert everything to RGBA?

tapir:
So my last commit, converts all non-RGBA images to RGBA (it's slow but setting cursor is not done ever frame so it shouldn't matter). No errors will occur since the type switch covers all the possibilities. Also now CreateCursor uses Go's image type instead of a custom image type. Much easier this way.
Tried and it works. Please take a look.

emidoots:
AFAIK the only resources you actually need to cleanup using the package are window's via `Destroy`.

I think in an effort of protecting users against potential memory leaks, we should attach a finalizer to `Cursor` and call `Cursor.Destroy` automatically. It would still be possible and useful to manually invoke `Destroy`, so it'd need to check that it hasn't yet been destroyed (a boolean flag protected by a mutex, for example).

tapir:
To me that's an overkill and not in the perimeters of a wrapper.

A cursor is hardly a memory leak anyway since applications will typically have only couple cursors which are used throughout the life of the application. Any developer that creates cursors very often, should also remember to destroy them afterwards.

I'm all about helping the developer (As in my recent non-RGBA to RGBA conversion code) but I'm also a believer that memory should be handled by the developer. Not us.

emidoots:
> memory should be handled by the developer. Not us.

Traditionally speaking, Go is garbage collected. The only instances which I know of in pure Go that you need to manage _resources_ are: ensuring goroutines exit, closing files, etc.

> A cursor is hardly a memory leak anyway since applications will typically have only couple cursors which are used throughout the life of the application.

The fact that there will only often be a few cursors in an application does not alleviate the fact that memory is being allocated and potentially never free'd (a _memory leak_).

I don't feel that it's a big enough issue to block on, though, so LGTM. :smiley:

dmitshur:
> I think in an effort of protecting users against potential memory leaks, we should attach a finalizer to `Cursor` and call `Cursor.Destroy` automatically. It would still be possible and useful to manually invoke `Destroy`, so it'd need to check that it hasn't yet been destroyed (a boolean flag protected by a mutex, for example).

Can you elaborate, I don't completely follow. What exactly do you mean by a finalizer? If it's something that runs when the app exits, wouldn't that be pointless, since `glfw.Terminate()` will destroy all resources anyway?

dmitshur:
> we should ... call `Cursor.Destroy` automatically

I might be missing some context, but I don't think we should be doing that at all. That would be very wrong. Only the user should call `Cursor.Destroy` since it's a part of the API.

emidoots:
@shurcooL [an object/runtime finalizer](http://golang.org/pkg/runtime/#SetFinalizer) -- e.g. attached to the `Cursor` struct such that when a `Cursor` is garbage collected, it's corresponding C memory would be free'd (if the user forgets to call `Cursor.Destroy` themselves)

dmitshur:
That is not neccessary since `glfw.Terminate()` does that.

> Any remaining cursors will be destroyed by glfwTerminate.

emidoots:
> That is not neccessary since glfw.Terminate() does that.

A finalizer runs when an object _is garbage collected_ -- not when the application exits (when `glfw.Terminate` is normally called).

Anyway, it's really not all that important.

emidoots:
Is it okay to merge this?

dmitshur:
Still reviewing.

input.go
```diff
@@ -344,6 +363,66 @@ func (w *Window) SetCursorPosition(xpos, ypos float64) error {
...
+// The cursor can be destroyed with Destroy. Any remaining cursors are destroyed by Terminate.
+//
+// The pixels are 32-bit little-endian RGBA, i.e. eight bits per channel. They are arranged
+// canonically as packed sequential rows, starting from the top-left corner.
+// 
+// All non-RGBA images will be converted to RGBA.
+//
+// The cursor hotspot is specified in pixels, relative to the upper-left corner of the cursor image.
+// Like all other coordinate systems in GLFW, the X-axis points to the right and the Y-axis points down.
+func CreateCursor(img image.Image, xhot, yhot int) (*Cursor, error) {
```
dmitshur, line 376: I'm still thinking whether it makes a better API/experience to accept `image.Image` instead of just `imageData []byte`...

tapir, line 376: byte array is not enough. you also need the width and height as in
GLFWimage struct. We can just create an Image type analog to GLFWimage and
be done with it. We can leave creating it to the developer or we can
include a helper CreateImage function that takes byte data width and height
as input.

emidoots, line 376: Wouldn't that struct end up looking pretty close to `*image.RGBA`?

dmitshur, line 376: > byte array is not enough. you also need the width and height as in
> GLFWimage struct.

I see.

tapir:
I like the idea of using Go types wherever possible. So image.Image makes sense I think.
I can understand the opposition against doing automatic color conversion though. Then again, in this case, it's fairly simple, and has no disadvantages I can think of. So, what do you guys say?

emidoots:
I personally like `image.Image` usage.

@shurcooL what are your thoughts here?

dmitshur:
Yeah, I think `image.Image` likely makes the most sense for the API.

tapir:
shall I merge then?

dmitshur:
Yeah, I think we can go with this. LGTM.

PR #119, opened 2015-01-19
Implement rest of the 3.1 features

tapir:


emidoots:
LGTM

dmitshur:
I see some minor issues, I'll fix them (unless you prefer me to tell you so you can fix yourself) by pushing onto this branch.

window.go
```diff
@@ -351,7 +361,19 @@ func (w *Window) SetSize(width, height int) error {
 func (w *Window) GetFramebufferSize() (width, height int, err error) {
 	var wi, h C.int
 	C.glfwGetFramebufferSize(w.data, &wi, &h)
-	return int(wi), int(h), fetchError()
+	return int(wi), int(h), fetchError() // FIXME: width and height are allowed to be NULL in C API
```
dmitshur: Can we fix this issue before merging the PR?

What is the problem exactly, and how should we best handle it?

tapir:
go on :thumbsup:

tapir:
oh sorry I didn't notice thr last email, I will think about it.

window.go
tapir: I really don't know in which case GLFW will return null. I'm guessing errors? In that case we can merge it for now. This is the idiomatic way and I would like to avoid doing multiple returns the C way as long as we don't criple functionality.

dmitshur: Are you talking about [this](http://www.glfw.org/docs/latest/group__window.html#ga0e2637a4161afb283f5300c7f94785c9)?

> This function retrieves the size, in pixels, of the framebuffer of the specified window. If you wish to retrieve the size of the window in screen coordinates, see glfwGetWindowSize.
> 
> Any or all of the size arguments may be NULL. If an error occurs, all non-NULL size arguments will be set to zero.
> 
> Parameters
>           [in]    window  The window whose framebuffer to query.
>            [out]  width   Where to store the width, in pixels, of the framebuffer, or NULL.
>           [out]   height  Where to store the height, in pixels, of the framebuffer, or NULL.

If so, that reads to me that the user of the C library has the option of providing nulls if he doesn't care about a certain value. But we're always providing non-nulls when calling the C func `glfwGetFramebufferSize`, so there will be no nulls to worry about.

Unless you meant something else, I don't think there's any issue here.

tapir: OK great. I've read it incorrectly. I'll remove the FIXMEs.

emidoots:
:+1:

PR #121, opened 2015-01-20
Return less errors; Fixes #118

### LINKED ISSUE #118: devel_glfw3.1: too many errors
@tapir [said](https://github.com/go-gl/glfw3/pull/113):

> The thing is now all the functions return an error and most of the errors are actually panics.
> Looks a bit overdone. Ideas?

emidoots:
For details about this solution see the documentation for `acceptError`, `panicError`, and issue #118.
### Present

I fairly confident this change will work appropriately, but I've only tested against my applications. As it's a rather large change, I would appreciate help reviewing / ensuring correctness and testing against any apps you might have.
### Future

`acceptError` will panic on unexpected error codes (ones that are not programmer errors or accepted by the calling function). This ensures that we will be able to catch such errors should we accidentally miss them in the future when updating the 'vendored' GLFW C sources.

I'm confident this change works with GLFW 3.1, but in the future we will need to carefully review changes to the C sources (specifically looking for new places errors might be generated) to ensure that we return them appropriately.

### DIFF
--- a/context.go
+++ b/context.go
@@ -12,32 +12,36 @@ import (
 // Originally GLFW 3 passes a null pointer to detach the context.
 // But since we're using receievers, DetachCurrentContext should
 // be used instead.
-func (w *Window) MakeContextCurrent() error {
+func (w *Window) MakeContextCurrent() {
 	C.glfwMakeContextCurrent(w.data)
-	return fetchError()
+	panicError()
+	return
 }
 
 // DetachCurrentContext detaches the current context.
-func DetachCurrentContext() error {
+func DetachCurrentContext() {
 	C.glfwMakeContextCurrent(nil)
-	return fetchError()
+	panicError()
+	return
 }
 
 // GetCurrentContext returns the window whose context is current.
-func GetCurrentContext() (*Window, error) {
+func GetCurrentContext() *Window {
 	w := C.glfwGetCurrentContext()
+	panicError()
 	if w == nil {
-		return nil, fetchError()
+		return nil
 	}
-	return windows.get(w), nil
+	return windows.get(w)
 }
 
 // SwapBuffers swaps the front and back buffers of the window. If the
 // swap interval is greater than zero, the GPU driver waits the specified number
 // of screen updates before swapping the buffers.
-func (w *Window) SwapBuffers() error {
+func (w *Window) SwapBuffers() {
 	C.glfwSwapBuffers(w.data)
-	return fetchError()
+	panicError()
+	return
 }
 
 // SwapInterval sets the swap interval for the current context, i.e. the number
@@ -54,9 +58,10 @@ func (w *Window) SwapBuffers() error {
 //
 // Some GPU drivers do not honor the requested swap interval, either because of
 // user settings that override the request or due to bugs in the driver.
-func SwapInterval(interval int) error {
+func SwapInterval(interval int) {
 	C.glfwSwapInterval(C.int(interval))
-	return fetchError()
+	panicError()
+	return
 }
 
 // ExtensionSupported returns whether the specified OpenGL or context creation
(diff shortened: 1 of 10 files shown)


tapir:
Everything looks in correct order.

> glfwCreateWindow
> glfwInit
> glfwIconifyWindow
> glfwRestoreWindow
> glfwGetClipboardString

Just in case let's take some time to make sure that above are the only functions with recoverable errors. Because if not we will have to break compatibility after the release. Or worse, live with an inconsistent API :p

dmitshur:
Sorry I've not yet had a chance to give this as much attention as I want to, but I just want to quickly comment now that from my brief look, this is a good change and it has my general support.

dmitshur:
Ok, I've looked the PR over and I think it looks good. I would say we could merge it (into `devel_glfw3.1` branch), and continue testing it with our own apps before official 3.1 PR merge into `master`.

However, before I say that, I want to bring up a point that I've discovered in the last hour when I was hacking on my [experimental browser backend](https://github.com/shurcooL/goglfw/commits/master) for glfw.

Namely, consider this func:

``` Go
func (w *Window) SetInputMode(mode InputMode, value int) error
```

After this PR, it becomes:

``` Go
func (w *Window) SetInputMode(mode InputMode, value int)
```

It seems that on all curent desktop backends (Win32, Cocoa, X11, Wayland, MIR, etc.) all input modes can never fail.

On the other hand, something like `IconifyWindow` or `GetClipboardString` can fail and it maintains an error return.

So, when implementing `SetInputMode` for the web, namely setting the cursor mode to one of `CursorNormal`, `CursorHidden`, or `CursorDisabled`, I realized that the first two will always succeed so no error return is needed, but `CursorDisabled` requires use of Pointer Lock API, which prompts the user for permission:

image

If the user says no, setting the cursor mode to `CursorDisabled` will fail, and ideally I would like for `SetInputMode` to be able to return an error code indicating denied permission, etc.

What if one of the desktop backends in the near future will have similar reasons why `SetInputMode` can sometimes fail, will we be able to easily add a return error type?

Just... wanted to share my findings, any thoughts? Basically, I'm starting to worry we're locking ourselves to current situation in terms of which funcs can return errors, and should anything unexpected happen and a new func needs to return errors, we have no way to do that except to break API. :/

tapir:
> Basically, I'm starting to worry we're locking ourselves to current situation in terms of which funcs can return errors, and should anything unexpected happen and a new func needs to return errors, we have no way to do that except to break API. :/

That worries me as well. Because of the way GLFW handles errors, it's easy for mainstream to return errors and not break API.

tapir:
Although, with `SetInputMode` you don't necessarily break API as the return value will just be ignored. It's a problem when there is multiple return values.

dmitshur:
> Although, with `SetInputMode` you don't necessarily break API as the return value will just be ignored. It's a problem when there is multiple return values.

That's true, but `SetInputMode` is just an example that I ran into recently. Another func may already have a return value, so adding error cannot be done seamlessly.

dmitshur:
The advantage of having errors in the API, while sometimes more verbose and superfluous, is that future possible error changes will not create extra work for us and users to update. It makes the API more future proof.

tapir:
I really wish GLFW was just using the C method of handling errors which is just return it. That would a lot easier for us but also for GLFW users as well I believe.

dmitshur:
> I really wish GLFW was just using the C method of handling errors which is just return it.

I don't see how that's relevant to us, since we achieved that by using channels (aside from making our API more different than the C one). But I don't think the C API can have that because C doesn't support multiple return values, so a func like `window = glfwCreateWindow(...)` cannot return both a window and error code.

emidoots:
The primary problem, as discussed largely already, is that GLFW does not document which functions exactly return which errors. Moreso, the C API is not as strictly tied to this documentation (if it existed).

#### If we return an error from every function

It's rather annoying and people will likely end up either ignoring them (which is fine for now, because not many functions generate errors, as seen here) or "just" printing them to stdout/stderr (_is this really any better than a error handler callback?_).

#### If we return an error for just those that actually generate them today

People will be more inclined to actually prepare to handle them appropriately and I think less buggy code will be written by the community at large.

#### Conclusion

I realize that this change isn't backwards compatible for future GLFW versions. If we merge this change, we're going to end up exposing more errors later (I have no doubt).

But I fear if we don't merge this, we risk creating a bad API that will encourage developers to write programs that don't handle errors -- just because we want to keep future compatibility.

dmitshur:
Whoops, I was swiping dust off my trackpad, and... Sorry.

tapir:
I'm inclined to agree with you. Let's merge this and test it.
Any new errors will have to wait for 3.2 then.

emidoots:
Hypothetically speaking: if GLFW uses semantic versioning, they should not add new errors for any existing functions as C code using it would also not be prepared to handle it.

PR #124, opened 2015-01-29
Allow use of OpenGL ES 2 on Desktop platforms with -tags=gles2

emidoots:
This allows for the explicitly choosing to use the OpenGL ES 2 client
instead of the standard OpenGL client on Desktop (386, amd64) platforms.

To explicitly choose the OpenGL ES 2 client, use the build tag: `gles2`.
- Additionally: Also choose to use OpenGL ES 2 by default when the `wayland` build tag is present. 
  - We still don't know if `-tags=wayland` builds correctly, but I think this betters the situation as Wayland primarily uses OpenGL ES 2 (not standard OpenGL).
  - If someone cares to test `wayland` later (after merge, or even after 3.1 release) I am willing to do the work to fix whatever might be broken.

### DIFF
--- a/build.go
+++ b/build.go
@@ -1,12 +1,13 @@
 package glfw3
 
 /*
-// Choose OpenGL client on 386 and amd64.
-#cgo 386 CFLAGS: -D_GLFW_USE_OPENGL
-#cgo amd64 CFLAGS: -D_GLFW_USE_OPENGL
+// Standard OpenGL client is used on 386 and amd64 architectures, except when
+// explicitly asked for gles2 or wayland.
+#cgo 386,!gles2,!wayland CFLAGS: -D_GLFW_USE_OPENGL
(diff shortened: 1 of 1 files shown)


emidoots:
FWIW I do have a real use-case for this: I am [experimenting with an OpenGL ES 2 backend](https://github.com/slimsag/gfx/commit/494b7845b9234ecf8b80602e4bd026b26ae8bb51) for something and it needs this change to run on Desktop.

In reality, you probably won't deploy a OpenGL ES 2 application on desktop anytime soon (not really as available compared to plain OpenGL) -- but this lets us test mobile OpenGL applications on desktop(!) instead of on a slow device or emulator.

dmitshur:
LGTM.

I can't verify this works (sadly, OpenGL ES is not available on OS X yet :( ), but I understand @slimsag has. Unless someone else who uses/depends on this functionality has comments, I think we can merge.

PR #125, opened 2015-02-18
update doc URL for new import paths

tapir:


dmitshur:
I think it's too early to do this. It makes sense to do everything in one PR, where the actual Go packages at `glfw/3.0/glfw` and `glfw/3.1/glfw` are populated. I can take care of all that.

PR #65, opened 2014-06-08
API changes and support for GLFW 3.1 features.

dmitshur:
This PR contains changes from the `devel` branch. That includes minor API changes that make the API easier to use and more idiomatic, as well as support for new features of GLFW 3.1. See changelog in README for a complete list of changes.

It is planned to be merged into master when GLFW 3.1 is released.

Closes #40.
Closes #8.
Closes #56.
Closes #106.

PR #126, opened 2015-02-22
Change import path according to early 2015 plan.

### LINKED ISSUE #123: Package name meta-issue, can we rename package to glfw?
The [example](https://github.com/go-gl/glfw3#example) in the README has:

``` Go
import (
    "fmt"
    glfw "github.com/go-gl/glfw3"
)
```

I always rename the `glfw3` package to `glfw` so that it can be used like `glfw.CreateWindow`. I think the majority of people do the same, but I wouldn't be surprised if I find some people forgetting to do that and having incompatible code like `glfw3.CreateWindow`. It's a trivial difference, but still.

In an ideal world, I would like to see:

`go-gl/glfw` - latest version of glfw
`go-gl/gl/...` - gl bindings (generated via glow)

`go-gl/glow` - only the glow binding generator itself

`go-gl/glfw-legacy` or `go-gl-legacy/glfw` - glfw 2.7 bindings.

That would be best for current users and new users who are just getting familiar with what is available under go-gl.

I don't think it's the best outcome to have legacy, outdated glfw 2.7 bindings using up the `github.com/go-gl/glfw` import path. Especially since this is Go, I see very little reason anyone would have to use older GLFW versions unless they have some legacy code they want to maintain; in all other cases you're better off with the latest stable GLFW version. This is not C++ where it's hard to update, just `go get -u`.

Of course, changing import paths is not easy, it has costs, and may not be possible. But I just wanted us to consider, can we do anything to improve the situation? What other people's thoughts?

dmitshur:
Closes #65.
Closes #123.

Move GLFW 3.0 package into v3.0/glfw subfolder.
Move GLFW 3.1 package into v3.1/glfw subfolder.
Rename both packages from glfw3 to glfw.

Once glfw repo is moved out of the way, this repo should first be renamed to glfw, then this PR can be merged.

This PR was made in the following 5 steps:
1. Starting with current master, I copied latest 3.0 bindings into v3.0/glfw subfolder. See commit bbf861bc46654eeee9fccad617bd2851ffb2e0fe.
2. I then rebased the latest `devel_glfw3.1` branch on top (resolving one merge conflict along the way). This got rid of merge commits, but otherwise should be identical changes to `devel_glfw3.1` branch.
3. After the rebase, I moved the latest 3.1 bindings into v3.1/glfw subfolder. See commit e3b81990b778b69f6593c284fa538f1341754f85.
4. I've made minimal updates to the README to point to new import paths. The README will likely need more thorough updating later on, but this is enough for first release. See commit aaf871d5344095d0360146535a2e1ed50220bbfe.
5. I've renamed both packages from glfw3 to glfw. (Discovered omission from testing.) See commit 5775b42f4cf8b47a49dfd5444939dce7ddca547c.

Please review. If I've made any mistakes or something isn't as expected, please point it out and I will correct it.


dmitshur:
I've tested both `v3.1/glfw` (with my own code) and `v3.0/glfw` (with cube and legacy_cube examples), and they both worked as expected.

tapir:
LGTM :+1:

emidoots:
Tested `v3.1/glfw` -- LGTM

Issue #127, opened 2015-02-23
Can not initialize GLFW without joystick support


tux21b:
I'm not quite sure what's the problem with the joystick configuration of my system, but the Go GLFW package panics with:

```
panic: PlatformError: Linux: Failed to watch for joystick connections in /dev/input: No space left on device
goroutine 16 [running]:
runtime.panic(0x501f80, 0xc20803e020)
    /home/christoph/go/src/pkg/runtime/panic.c:279 +0xf5
github.com/go-gl/glfw/v3.1/glfw.acceptError(0x7f769b90ae98, 0x1, 0x1, 0x0, 0x0)
    /home/christoph/gopath/src/github.com/go-gl/glfw/v3.1/glfw/error.go:164 +0x13d
github.com/go-gl/glfw/v3.1/glfw.Init(0x0, 0x0)
    /home/christoph/gopath/src/github.com/go-gl/glfw/v3.1/glfw/glfw.go:33 +0x70
```

The error itself is triggered at `glfw/src/linux_joystick.c:_glfwInitJoysticks():213` which contains a comment in the following line saying: `// Continue without device connection notifications`. Therefore, I expect everything (except discovering new joysticks) to work. Unfortunately, the GLFW Go binding is way more strict and throws a panic in this case.

tapir:
Can you post some example code that triggers this?

Edit: We've made platform errors panic in purpose because they are supposed to be not recoverable. I will check.

tux21b:
> Edit: We've made platform errors panic in purpose because they are supposed to be not recoverable. I will check.

I am not sure if that's the right thing, especially for those errors that are explicitly non-fatal according to the GLFW source. There are quite a lot of such ignore / resume without ... comments in the GLFW C source.

Here is a full example, but I doubt that it will be of much use. I'm still not sure why inotify isn't working for /dev/input on this system (there is plenty of available disk space of course):

```
package main

import (
    "github.com/go-gl/glfw/v3.1/glfw"
    "log"
    "runtime"
)

func init() {
    runtime.LockOSThread()
}

func main() {
    log.Println("GLFW version:", glfw.GetVersionString())

    err := glfw.Init()
    if err != nil {
        log.Fatal(err)
    }
    defer glfw.Terminate()
}
```

Output:

```
[christoph@tuxmobil ~]$ go run glfw_bug.go 
2015/02/24 10:08:03 GLFW version: 3.1.0 X11 GLX glXGetProcAddressARB clock_gettime /dev/js
[...]

PR #130, opened 2015-03-10
v3.1: print PlatformErrors instead of panicking on them

emidoots:
For details please see the discussion in the issue.

Workaround for #127


dmitshur:
LGTM because it's an improvement over current behavior, but I anticipate/hope for further improvements in this area. We can do that on an as-needed basis once issues are opened, though.

PR #133, opened 2015-03-20
v3.1/glfw: Update to GLFW 3.1.1.

### LINKED ISSUE #132: GLFW 3.1.1
We should look into changing `go-gl/glfw/v3.1` to be v3.1.1 or something. It was released recently and adds a workaround for a critical bug that renders GLFW useless on most Linux window managers in the wild today.

@shurcooL any thoughts regarding this?

dmitshur:
A new patch version of GLFW has been released:

https://github.com/glfw/glfw/releases - 3.1.1 - glfw/glfw@29b4011

This PR updates the `v3.1/glfw` to use it. It was created by following the [usual procedure](https://github.com/go-gl/glfw/wiki/Development#updating-glfw-c-source).

Closes #132.


tapir:
LGTM

emidoots:
Awesome work!

It looks like the generated errors have changed / got shifted around a bit -- so I will want to triple-check to make sure nothing will break (I'll have time to review later this weekend).

dmitshur:
Sure, there's no need to rush on merging this.

I've created it because it was quick and easy, and so I (and others) can start testing this branch. I'm running it locally so I'll see if I spot any issues over the next day or two.

dmitshur:
There are two reports of this PR fixing glfwCreateWindow blocking problems on Ubuntu 14.04, see https://github.com/google/gxui/issues/35#issuecomment-84352668 and https://github.com/google/gxui/issues/35#issuecomment-84370899.

emidoots:
v3.1.1 does mix around some of the error codes, so in the event of those errors there might be a unwarranted panic. We can fix those as they appear in the wild.

I'll also see about contributing some better GLFW (C lib) documentation on which functions generate which errors, or maybe I'll write a ugly/hacky tool to automate the process.

For now, LGTM. Lets merge this in as it fixes a real-world problem users are facing today.

EDIT: grepping the diff for `GLFW_VERSION_UNAVAILABLE` will show the ones I'm primarily concerned about (those should probably be actual Go `error`s).

Issue #140, opened 2015-07-19
cannot find -lXxf86vm


pebbe:
Installation fails:

```
(peter) ~ go get github.com/go-gl/glfw/v3.1/glfw
# github.com/go-gl/glfw/v3.1/glfw
/usr/bin/ld: cannot find -lXxf86vm
collect2: ld returned 1 exit status
(peter) ~ ll /usr/lib/libXxf86vm*
lrwxrwxrwx 1 root root    19 jun 10  2013 /usr/lib/libXxf86vm.so.1 -> libXxf86vm.so.1.0.0
-rw-r--r-- 1 root root 19808 mei 30  2013 /usr/lib/libXxf86vm.so.1.0.0
```

go version go1.4.2 linux/amd64
Debian 6.0

dmitshur:
This is related to #138. We need to improve the README.

First, can you try a few things:
1. `go get -u github.com/go-gl/glfw/v3.1/glfw` Use the -u flag for go get.
2. If the above doesn't work, try `sudo apt-get install libgl1-mesa-dev xorg-dev`. What does that say? Afterwards, try the go get -u step above.

Let me know if that works.

pebbe:
After installing `xorg-dev`, installation of glfw succeeds. And I can build the examples. But when I run an example I get:

```
failed to initialize glfw: APIUnavailable: GLX: GLX version 1.3 is required
```

Searching for packages with glx in their name, I get plenty. I wouldn't now which to install.

dmitshur:
I just booted up a fresh Ubuntu (it's what I had) VM and after installing Go, I did this one step:

```
sudo apt-get install libgl1-mesa-dev xorg-dev
```

Afterwards `go get -u github.com/go-gl/glfw/v3.1/glfw` worked (as well as building and running [examples](https://github.com/go-gl/examples)).

FWIW, I looked at where the `libXxf86vm*` files ended up, and it was elsewhere. (But I see you've already commented above that it worked for you after getting `xorg-dev`.)

dmitshur:
A GLX version issue is something I expect on when the system is strangely configured and doesn't support OpenGL (e.g., some old VM or really old hardware), or it might be a graphics card drive issue. I'm not sure if it's just a missing library...

However, this might be a GLFW (the C library) issue possibly (but more likely it's something weird with your setup that can be avoided). Take a look at https://github.com/glfw/glfw/issues/522 and see if it's relevant to you.

[...]

PR #141, opened 2015-07-19
README: Update installation steps, mention dependencies, remove older comments.

### LINKED ISSUE #138: Dependencies not clearly documented
I just tried to compile a program using glfw 3.1, and hit this error:

```
$ go build
# github.com/go-gl/glfw/v3.1/glfw
In file included from ../../.local/src/github.com/go-gl/glfw/v3.1/glfw/native_linbsd.go:8:0:
glfw/include/GLFW/glfw3native.h:88:36: fatal error: X11/extensions/Xrandr.h: No such file or directory
  #include <X11/extensions/Xrandr.h>
                                    ^
```

I happen to know that I should install the build dependencies for libglfw, so I can run this:

```
$ sudo apt-get build-dep libglfw-dev
```

To solve the problem. But it isn't documented anywhere. What should we document exactly?

dmitshur:
Use `-u` flag for go get command to ensure latest version is installed (latest versions tend to exist in order to fix bugs and issues).

Mention that dependencies of GLFW (C library) need to be installed (but not GLFW itself, that is built as part of Go package).

Remove old comments like "we're breaking the API" as that's a constant now; the changelog remains below. Also remove comment about adding examples since the README in examples repo does a better job of communicating how to best contribute there. Change title from "bindings" since it's now a more complete Go package, not just bindings. Update Go issue 8811 link to current location.

Use a godoc badge for the documentation link.

Fixes #138 (and related to #140).

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,20 @@
-# Go Bindings for GLFW 3.1
-
-* **ATTENTION:** As of GLFW 3.1 we break API. See Changelog below.
-* See [here](http://godoc.org/github.com/go-gl/glfw/v3.1/glfw) for documentation.
-* You can help by submitting examples to [go-gl/examples](http://github.com/go-gl/examples).
+# GLFW 3.1 for Go [![GoDoc](https://godoc.org/github.com/go-gl/glfw/v3.1/glfw?status.svg)](https://godoc.org/github.com/go-gl/glfw/v3.1/glfw)
 
 ## Installation
 
-* Installation is easy, just `go get github.com/go-gl/glfw/v3.1/glfw` and be done (*GLFW sources are included so you don't have to build GLFW on your own*)!
-* Go 1.4 is required on Windows (otherwise you must use MinGW v4.8.1 exactly, see [Go issue 8811](https://code.google.com/p/go/issues/detail?id=8811)).
+* GLFW C library source is included and built automatically as part of the Go package. But you need to make sure you have dependencies of GLFW:
+	* On OS X, you need Xcode for required headers and libraries.
+	* On Linux, you need `libgl1-mesa-dev` and `xorg-dev` packages.
+	* See [here](http://www.glfw.org/docs/latest/compile.html#compile_deps) for full details.
+* Go 1.4+ is required on Windows (otherwise you must use MinGW v4.8.1 exactly, see [Go issue 8811](https://github.com/golang/go/issues/8811)).
+
+```
+go get -u github.com/go-gl/glfw/v3.1/glfw
+```
 
 ## Usage
 
-```go
+```Go
 package main
 
 import (
@@ -40,7 +43,7 @@ func main() {
 	window.MakeContextCurrent()
 
 	for !window.ShouldClose() {
-		// Do OpenGL stuff
+		// Do OpenGL stuff.
 		window.SwapBuffers()
 		glfw.PollEvents()
 	}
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,17 +1,20 @@
...
-* See [here](http://godoc.org/github.com/go-gl/glfw/v3.1/glfw) for documentation.
-* You can help by submitting examples to [go-gl/examples](http://github.com/go-gl/examples).
+# GLFW 3.1 for Go [![GoDoc](https://godoc.org/github.com/go-gl/glfw/v3.1/glfw?status.svg)](https://godoc.org/github.com/go-gl/glfw/v3.1/glfw)
 
 ## Installation
 
-* Installation is easy, just `go get github.com/go-gl/glfw/v3.1/glfw` and be done (*GLFW sources are included so you don't have to build GLFW on your own*)!
-* Go 1.4 is required on Windows (otherwise you must use MinGW v4.8.1 exactly, see [Go issue 8811](https://code.google.com/p/go/issues/detail?id=8811)).
+* GLFW C library source is included and built automatically as part of the Go package. But you need to make sure you have dependencies of GLFW:
+	* On OS X, you need Xcode for required headers and libraries.
```
dmitshur, line 6: While I've verified (via a fresh Ubuntu VM) the Linux dependencies are indeed mandatory, I haven't been able to verify the that Xcode is absolutely needed for OS X. It probably is. I copied it from the GLFW C library documentation:

> Xcode comes with all necessary tools except for CMake. The required headers and libraries are included in the core OS X frameworks. Xcode can be downloaded from the Mac App Store or from the ADC Member Center.

So it's probably correct. But I've always had Xcode installed on my Macs, so it's hard to verify. If someone else can, that'd be great.

README.md
```diff
@@ -1,17 +1,20 @@
...
+	* On Linux, you need `libgl1-mesa-dev` and `xorg-dev` packages.
```
tapir: xorg-dev is a meta package for debian/ubuntu derivatives.
Instead of "on linux" I would prefer "on debian/ubuntu derivatives"
Here is the list of packages that xorg-dev installs: https://packages.debian.org/sid/x11/xorg-dev

most of them are probably not necessary

dmitshur: I know, but it's what GLFW suggests.

From http://www.glfw.org/docs/latest/compile.html#compile_deps:

> ### Dependencies for Linux and X11
> 
> To compile GLFW for X11, you need to have the X11 and OpenGL header packages installed, as well as the basic development tools like GCC and make. For example, on Ubuntu and other distributions based on Debian GNU/Linux, you need to install the **`xorg-dev`** and `libglu1-mesa-dev` packages. ...

tapir: This is precisely what I suggest. We should add the keywords Debian and/or Ubuntu somewhere there to avoid confusion for people that use other distros.

dmitshur: How about:

> On Ubuntu/Debian-like Linux distributions, you need ...

dmitshur: Alternatively, instead of trying to summarize (inaccurately) the dependencies, we can just omit the short versions and simply link to http://www.glfw.org/docs/latest/compile.html#compile_deps. That way people get full details on the dependencies.

The main reason I wanted to avoid that is because that page is large and includes many irrelevant things (other sections), so users may end up inadvertently following additional steps there that make no sense.

dmitshur:
How's that look?

tapir:
:+1:

PR #149, opened 2015-11-02
Add Travis CI for glfw v3.1 Go package.

dmitshur:
I expect it to fail at this time due to https://github.com/go-gl/glfw/pull/148#issuecomment-152880115.

I'll try to resolve that on this branch.

**Edit:** 6a42f3c appears to fix the issue. It needs to be reviewed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: go
+go:
+  - 1.5
+before_install:
+  - sudo apt-get update
+  - sudo apt-get install -qq libgl1-mesa-dev xorg-dev
+install:
+  - go get golang.org/x/tools/cmd/vet
+script:
+  - go get -t -v ./v3.1/...
+  - diff -u <(echo -n) <(gofmt -d .)
+  - go tool vet .
+  - go test -v -race ./v3.1/...
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,6 @@ install:
   - go get golang.org/x/tools/cmd/vet
 script:
   - go get -t -v ./v3.1/...
-  - diff -u <(echo -n) <(gofmt -d .)
+  - diff -u <(echo -n) <(gofmt -d -s .)
   - go tool vet .
   - go test -v -race ./v3.1/...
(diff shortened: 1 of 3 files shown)


dmitshur:
This is now ready for review and merging if it's good, PTAL.

.travis.yml
```diff
@@ -0,0 +1,13 @@
...
+go:
+  - 1.5
+before_install:
+  - sudo apt-get update
+  - sudo apt-get install -qq libgl1-mesa-dev xorg-dev
+install:
+  - go get golang.org/x/tools/cmd/vet
+script:
+  - go get -t -v ./v3.1/...
+  - diff -u <(echo -n) <(gofmt -d -s .)
```
emidoots, line 11: What does this do? Please add a comment describing.

emidoots:
One comment, but LGTM.

Note that its because Travis CI uses such a dated Ubuntu version that -lrt is needed. This is also why I and others didn't experience this issue. More info:   http://stackoverflow.com/a/32649327

.travis.yml
dmitshur, line 11: I thought this was a very common idiom, that's why there's no comment. E.g., see [here](https://github.com/go-gl/examples/blob/26d37bbf6265a393404a126a43ac69930370e1fb/.travis.yml), [here](https://github.com/sourcegraph/go-diff/blob/7eefa0faf57871a70d7a2c368a454593f07797bc/.travis.yml#L18) and [here](https://github.com/shurcooL/goe/blob/3bcc0e08dda8182005ab103f77e7bf8749d1cf7e/.travis.yml#L8).

It tests that there are no `gofmt` issues.

More specifically, it prints a unified diff (for ease of reading) between empty string (`echo -n`) and the result of running `gofmt -d -s .`, which runs `gofmt` on current directory (and [therefore](https://golang.org/cmd/gofmt/) all .go files inside, recursively) and prints any `-d`ifferences from correct `gofmt` formatting. It also checks for `-s`implify opportunities.

So, if there are no `gofmt` issues, the diff is non-existent and exit code from `diff` is 0, build passes. If there are non-zero differences, then `diff` returns non-zero exit code. Running `gofmt -d` alone does not return a non-zero exit code if there are formatting issues.

dmitshur:
> Note that its because Travis CI uses such a dated Ubuntu version that -lrt is needed. This is also why I and others didn't experience this issue. More info:   http://stackoverflow.com/a/32649327

I see. I kinda suspected that.

Is there any reasonable way we can limit this change to Travis, or is it okay/reasonable to include this for all Linux distributions?

I think we must support Travis builds so that other packages that import `glfw` can build successfully and have their tests pass. But if it's possible to do so in a better way, then we can do that.

I'll merge this so that importers of this package can pass tests. But if there are improvement/enhancement suggestions, let's work on implementing that as a followup.

.travis.yml
dmitshur, line 11: In other words, the reason I think this doesn't deserve a comment is because it is the minimum (known to me) set of commands to check for `gofmt` issues. If this line can be made any simpler, we should do that.

Since it mentions `gofmt` which should be a good hint of what it does, and is the simplest possible command, I don't think a comment is needed.

dmitshur, line 11: One more thing, the `diff <(foo) <(bar)` pattern is called [process substitution](https://en.wikipedia.org/wiki/Process_substitution). Bash supports it.

PR #153, opened 2015-12-13
Travis: Update.

dmitshur:
Use container-based builds.

Build for tip (with allowed failure and fast finish).


luna-duclos:
LGTM

Issue #156, opened 2016-04-12
glfw.CreateCursor causes a panic


rbudzko:
Hi,

glfw.CreateCursor causes a panic, due to improvement introduced in (I believe) golang 1.6. We are encountering "Go pointer to Go pointer" passed to CGO stack. We are talking about img_c structure in input.go and more specifically img_c.pixels around line ~390.

Problem can be solved by for example improvement like this:

c_pix, destructor := gl.Strs(string([]byte(m.Pix)) + "\x00")
img_c.pixels =  (_C.uchar)(_c_pix)

I'm more then happy to prepare pull request, but first I'd like to confirm if importing gl to glfw is appropriate. Additionally, destructor can be called at the end of glfw.CreateCursor, because C.glfwCreateCursor seems to copy the buffer.

Robert Budźko

dmitshur:
Hi @rbudzko,

Thanks for reporting the issue. Your diagnosis sounds spot on to me.

> I'm more then happy to prepare pull request, but first I'd like to confirm if importing gl to glfw is appropriate.

In this case, it makes the most sense to simply copy the relevant snippet of code rather than importing all of `gl` for it.

There are many `gl` libraries, and unless the user imports the exact same one, it'll end up pulling in a large dependency unnecessarily.

Instead, feel free to just copy the `Strs` code from `gl` here. It's analogous to how each `gl` version package has its own. You should make it unexported since it's not going to be a part of `glfw`'s API.

Thanks for making a PR for this, I'll be happy to review it (or anyone else)!

tapir:
I recommend _util.go_ for that function and also could you make a PR for devel branch v3.2 as well please.

Thank you.

rbudzko:
@tapir thanks for a guidance! Prepared PR.

PR #157, opened 2016-04-13
#156 - CreateCursor copies pixels off Go space and manually deallocates it after cursor is created.

rbudzko:
Noticed that while v3.1 passes test (in my internal application), v3.2 is failing on missing c symbols:

> could not determine kind of name for C.glfwGetTimerFrequency
> could not determine kind of name for C.glfwGetTimerValue

I've intentionally ignored this fact since HEAD~1 also fails and it's a development branch.

### DIFF
--- a/v3.1/glfw/input.go
+++ b/v3.1/glfw/input.go
@@ -381,23 +381,29 @@ func (w *Window) SetCursorPos(xpos, ypos float64) {
 // Like all other coordinate systems in GLFW, the X-axis points to the right and the Y-axis points down.
 func CreateCursor(img image.Image, xhot, yhot int) *Cursor {
 	var img_c C.GLFWimage
-	var pixPointer *uint8
+	var pixels []uint8
 	b := img.Bounds()
 
 	switch img := img.(type) {
 	case *image.RGBA:
-		pixPointer = &img.Pix[0]
+		pixels = img.Pix
 	default:
 		m := image.NewRGBA(image.Rect(0, 0, b.Dx(), b.Dy()))
 		draw.Draw(m, m.Bounds(), img, b.Min, draw.Src)
-		pixPointer = &m.Pix[0]
+		pixels = m.Pix
 	}
+
+	pix, free := strs(string(pixels) + "\x00")
 
 	img_c.width = C.int(b.Dx())
 	img_c.height = C.int(b.Dy())
-	img_c.pixels = (*C.uchar)(pixPointer)
+	img_c.pixels = (*C.uchar)(*pix)
+
 	c := C.glfwCreateCursor(&img_c, C.int(xhot), C.int(yhot))
+
+	free()
 	panicError()
+
 	return &Cursor{c}
 }
 
--- a/v3.1/glfw/input.go
+++ b/v3.1/glfw/input.go
@@ -393,11 +393,11 @@ func CreateCursor(img image.Image, xhot, yhot int) *Cursor {
 		pixels = m.Pix
 	}
 
-	pix, free := strs(string(pixels) + "\x00")
+	pix, free := bytes(pixels)
 
 	img_c.width = C.int(b.Dx())
 	img_c.height = C.int(b.Dy())
-	img_c.pixels = (*C.uchar)(*pix)
+	img_c.pixels = (*C.uchar)(pix)
 
 	c := C.glfwCreateCursor(&img_c, C.int(xhot), C.int(yhot))
 
--- a/v3.1/glfw/util.go
+++ b/v3.1/glfw/util.go
@@ -1,11 +1,53 @@
 package glfw
 
+//#include <stdlib.h>
 //#include "glfw/include/GLFW/glfw3.h"
 import "C"
+
+import (
+	"reflect"
+	"unsafe"
+)
 
 func glfwbool(b C.int) bool {
 	if b == C.GL_TRUE {
 		return true
 	}
 	return false
+}
+
+// strs takes a list of Go strings (with or without null-termination) and
+// returns their C counterpart.
+//
+// The returned free function must be called once you are done using the strings
+// in order to free the memory.
+//
+// If no strings are provided as a parameter this function will panic.
+func strs(strs ...string) (cstrs **uint8, free func()) {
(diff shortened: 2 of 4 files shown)


v3.2/glfw/input.go
```diff
@@ -390,23 +390,29 @@ func (w *Window) SetCursorPos(xpos, ypos float64) {
...
-		pixPointer = &img.Pix[0]
+		pixels = img.Pix
 	default:
 		m := image.NewRGBA(image.Rect(0, 0, b.Dx(), b.Dy()))
 		draw.Draw(m, m.Bounds(), img, b.Min, draw.Src)
-		pixPointer = &m.Pix[0]
+		pixels = m.Pix
 	}
 
+	pixPtr_c, pixFree_c := str(string(pixels) + "\x00")
```
dmitshur: What is `str`? Shouldn't it be `strs`, the new func you've added? I'm guessing it's a typo, right?

v3.1/glfw/util.go
```diff
@@ -1,11 +1,53 @@
...
+)
+
 func glfwbool(b C.int) bool {
 	if b == C.GL_TRUE {
 		return true
 	}
 	return false
 }
+
+// Strs takes a list of Go strings (with or without null-termination) and
```
dmitshur: Since it's no longer exported, it's called `strs`, not `Strs`. The documentation should be updated too.

v3.1/glfw/input.go
```diff
@@ -381,23 +381,29 @@ func (w *Window) SetCursorPos(xpos, ypos float64) {
...
-		pixPointer = &img.Pix[0]
+		pixels = img.Pix
 	default:
 		m := image.NewRGBA(image.Rect(0, 0, b.Dx(), b.Dy()))
 		draw.Draw(m, m.Bounds(), img, b.Min, draw.Src)
-		pixPointer = &m.Pix[0]
+		pixels = m.Pix
 	}
 
+	pixPtr_c, pixFree_c := strs(string(pixels) + "\x00")
```
dmitshur: These names are not idiomatic to Go (see https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps) and feel too long. They're used pretty much immediately. What do you think about just calling them `pix` and `free`?

rbudzko:
@shurcooL thank you for pointing out those mistakes. I've adjusted the code.

dmitshur:
Sorry I didn't spot this earlier, but `strs` is not the most appropriate helper to use for this. You want to copy a single contiguous byte slice, not multiple null-terminated C strings.

This is a great start but we can simplify the `strs` code significantly, I think. /cc @slimsag Do you remember if there's a simpler byte copier snippet already available anywhere?

rbudzko:
@shurcooL your are right. I've copied strs from gl code base a little bit too blindly. Adjusted implementation, so it's more like non-exported and byte oriented implementation.

dmitshur:
This is looking much better, thanks. I have some more comments to improve `bytes`.

v3.1/glfw/util.go
```diff
@@ -1,11 +1,36 @@
...
 	if b == C.GL_TRUE {
 		return true
 	}
 	return false
 }
+
+func bytes(origin []byte) (pointer *uint8, free func()) {
+	if len(origin) == 0 {
+		panic("Slice of bytes cannot be empty.")
+	}
```
dmitshur, line 24: I don't see the point of panicking if length of `origin` is 0. In `gl.Strs`, it was needed to handle invalid API usage of providing no strings. But I'm not sure that having it here for `bytes` makes sense. If there's a way in which it's helpful, can you explain it?

tapir:
I've also commented about the same thing yesterday (though now I see it was only left as a note not comment) I believe panic is not needed.

v3.1/glfw/util.go
```diff
@@ -1,11 +1,36 @@
...
+
+	l := len(origin)
+	d := C.malloc(C.size_t(l))
+
+	ds := *(*[]byte)(unsafe.Pointer(&reflect.SliceHeader{
```
dmitshur: At first I found it hard to understand what `d` and `ds` stood for, then I realized it was a short version of `data` and `dataSlice` from original `gl.Strs` code. I think the original names were more readable, let's stick with those, unless you feel strongly about it.

v3.1/glfw/util.go
```diff
@@ -1,11 +1,36 @@
...
+		Data: uintptr(d),
+		Len:  l,
+		Cap:  l,
+	}))
+
+	copy(ds[0:l], origin[:])
```
dmitshur: Isn't this the same as `copy(ds, origin)`? If so, it's better to write that, since it's simpler.

v3.1/glfw/util.go
```diff
@@ -1,11 +1,36 @@
...
+
+	return (*uint8)(unsafe.Pointer(&ds[0])), func() { C.free(d) }
```
dmitshur: Similarly, isn't this the same as `return &ds[0], func() { C.free(d) }`? If so, let's use the simpler version.

v3.1/glfw/util.go
```diff
@@ -1,11 +1,36 @@
...
 	}
 	return false
 }
+
+func bytes(origin []byte) (pointer *uint8, free func()) {
+	if len(origin) == 0 {
+		panic("Slice of bytes cannot be empty.")
+	}
+
+	l := len(origin)
```
dmitshur: Nitpick, but I like `n` better than `l` here. It's more consistent with most similar Go code I've seen.

Or maybe even avoid it altogether, it's only used in 2 places.

rbudzko:
Wasn't aware that glfw tolerates NULL if width or height is 0.

dmitshur:
LGTM. Thank you for reporting and fixing the issue, @rbudzko!

I'll merge this both into `devel` and into `master`.

dmitshur:
Merged as https://github.com/go-gl/glfw/commit/d365d52b45af0a670284b1cf22f8d90d5de94513 and https://github.com/go-gl/glfw/commit/e08e51ed34fa8be1bab154a41910d15c6e9202ac.

PR #169, opened 2016-08-19
Travis: Use current stable and previous Go releases.

dmitshur:
We were testing a really old version of Go 1.5.4, and tip (with allowed failure).

Update to current release of Go (1.7) and previous version (1.6).

This seems reasonable, right? I don't think we should aim to support much more than that, we have very limited resources.


tapir:
:+1:

Issue #172, opened 2016-09-23
Expose Window.data as *C.GLFWwindow


xlab:
Hi, was going to use GLFW to init the window and then pass it to the next (GUI) layer that expects a window handle for the further operation.

``` go
type Window struct {
    data *C.GLFWwindow
    // other fields
}
```

It seems that there is currently no way to get this handle from the outside. That blocks me from using the package. :(

dmitshur:
Have you seen [`glfw.Window.GetX11Window`](https://godoc.org/github.com/go-gl/glfw/v3.2/glfw#Window.GetX11Window) and similar funcs?

There's a different one for each OS. See:
- https://github.com/go-gl/glfw/blob/master/v3.2/glfw/native_darwin.go
- https://github.com/go-gl/glfw/blob/master/v3.2/glfw/native_linbsd.go
- https://github.com/go-gl/glfw/blob/master/v3.2/glfw/native_windows.go

xlab:
Thing is that I don't want to get `ret := uintptr(C.workaround_glfwGetCocoaWindow(w.data))` but the `w.data` itself, which is `*C.GLFWwindow`, not a platform-native window handle. I have a [GUI layer](https://github.com/golang-ui/nuklear) that wants `*C.GLFWwindow` :)

emidoots:
I'd be fine with exposing a new method in `https://github.com/go-gl/glfw/blob/master/v3.2/glfw/native.go` like:

```
// GetGLFWWindow returns a *C.GLFWwindow (i.e. the GLFW window itself). This can be used for
// passing the GLFW window handle to external C libraries.
func (w *Window) GetGLFWWindow() uintptr {
    ... return w.data
}
```

Want to send a PR?

xlab:
Cool, I'll propose tomorrow. One more question though, what's up with 'uintptr' instead of 'unsafe.Pointer'?

dmitshur:
Does GLFW (the C library, /cc @elmindreda) have an equivalent to the proposed `GetGLFWWindow()`? If not, how does it get by without one?

I'd be a little wary of adding methods not present in the C library. At least, I'd like to really understand why we're doing that, before doing it.

Also, style-wise, it should be `GLFWWindow()`, not `GetGLFWWindow()`. See https://golang.org/doc/effective_go.html#Getters:

> but it's neither idiomatic nor necessary to put Get into the getter's name.

emidoots:
[...]

PR #173, opened 2016-09-28
Expose *C.GLFWwindow ref. Updates #172.

xlab:


tapir:
I would rather have this where Window is defined.

tapir:
Also we can lose "Get". The go way is to use the name directly for getters.

xlab:
@tapir please see the discussion in #172 
The file name has been suggested to me and we decided that `Get` should stay to keep the API consistent.

tapir:
@xlab those functions are not getters in the sense we're talking about. They directly correspond to glfw methods with the same names. However in this case we're trying to get a reference to a private value inside Window. So it should be just GFLWWindow()

tapir:
We can discuss it further if you like @slimsag @shurcooL

dmitshur:
I was originally convinced by @slimsag's comment:

> > Also, style-wise, it should be GLFWWindow(), not GetGLFWWindow(). See https://golang.org/doc/effective_go.html#Getters.
> 
> @shurcooL I am well aware, but suggested the `Get` prefix because that's what we use in all three files you linked (`GetCocoaWindow`, `GetX11Window`, etc).

But now I've remembered what's going on. We kept those `Get` in order to have names that match the GLFW C API names:

http://www.glfw.org/docs/latest/group__native.html

I agree with @tapir now, since this is a pure Go construct, there's no C API to follow, we should use Go style and drop the "Get".

Also agree it doesn't deserve a new file.

xlab:
@tapir @shurcooL yeah, I like this better.

Issue #175, opened 2016-10-30
/usr/bin/ld: cannot find -lXi which library is this?


choleraehyq:
My OS is fedora 24, I've already read the dependencies and installed:

```
mesa-libGL-devel
libX11-devel
libXinerama-devel
libXrandr-devel
libXcursor-devel
```

and the corresponding packages without `-devel` suffix.
But when I ran `go get github.com/go-gl/glfw/v3.1/glfw` an error occured:

```
/usr/bin/ld: cannot find -lXi
collect2: error: ld returned 1 exit status
```

And which package should I install?

choleraehyq:
Sorry, lt's libXi-devel, my fault.

dmitshur:
No worries.

The [installation section](https://github.com/go-gl/glfw#installation) in README mentions packages `libgl1-mesa-dev` and `xorg-dev`. Did you see that? Are they available on Fedora 24, or not?

If not, perhaps we should add these to the README so other Fedora users have better documentation to work with. Do you think that'd be helpful?

choleraehyq:
Thanks for your reply.
I've already read the README carefully, but these two packages are only available on Ubuntu/Debian. I think it's very helpful if you add fedora dependencies to the README.  After all, CentOS/Redhat/Fedora-like linux distributions are also widely used.

dmitshur:
> I think it's very helpful if you add fedora dependencies to the README. After all, CentOS/Redhat/Fedora-like linux distributions are also widely used.

Agreed, we should do that.

Would you like to make a PR to do that? Just add a new paragraph that begins with something like "On CentOS/Redhat/Fedora-like Linux distributions, ...". Sort them by most popular first.

It's hard for me since I don't know anything about those systems, but I hope it's easy for you. Thanks!

choleraehyq:
OK, I'll work on this.

PR #176, opened 2016-10-31
Add package dependencies on CentOS/Fedora-like distribution to README

choleraehyq:
I've already tested this dependencies list on CentOS 7 and Fedora 24. Redhat is not free so I cannot test on it right now.

@shurcooL Please review it, thanks!

corresponding issue: #175


dmitshur:
I've made some formatting changes, otherwise LGTM.

Issue #179, opened 2016-12-29
Unable to go get glfw v3.2


codephobia:
Trying to use the following command ends up just returning an error, but works for v3.1.

Command:
`go get -u github.com/go-gl/glfw/v3.2/glfw`

Error:
```
# github.com/go-gl/glfw/v3.2/glfw
In file included from ./glfw/src/internal.h:167:0,
                 from ./glfw/src/context.c:28,
                 from ..\..\go-gl\glfw\v3.2\glfw\c_glfw.go:4:
./glfw/src/win32_platform.h:70:20: fatal error: dinput.h: No such file or directory
compilation terminated.
```

dmitshur:
This looks very similar to #167. Can you check if that's it?

codephobia:
The file seems to be there...

`C:\gopath\src\github.com\go-gl\glfw\v3.2\glfw\glfw\deps\mingw\dinput.h`

codephobia:
I moved the 3 files from the deps folder `go-gl\glfw\v3.2\glfw\glfw\deps\mingw\` to the main folder `go-gl\glfw\v3.2\glfw` and then it was able to build with v3.2.

dmitshur:
Oh, I didn't realize we had a copy of `dinput.h` in the tree. Perhaps we should update the include directories to look there.

I've never looked at this because I don't use Windows at this time.

tapir:
Hmm, what gcc and go environment do you use? On my system (win7 64bit, msys2-64) I don't get such error.

codephobia:
Win10 64bit, go version go1.7.3 windows/amd64, gcc 5.3.0-3 installed using MinGW Installation Manager.

tapir:
I wonder if a `-Iglfw/deps/mingw` in build.go work?

codephobia:
I tried to build my code on another machine, and it seems I am getting this error again, but this time I can't seem to be able to resolve it.

codephobia:
Oh, it looks like my go version was a little out of date. Updating to latest version resolved the issue.

dmitshur:
> but this time I can't seem to be able to resolve it.

> Updating to latest version resolved the issue.

Do you mean the _original_ issue is completely resolved now, or the _new_ one you just ran into recently?

Did you still have to move some files in the repo manually for it to work?

nxtlvlgg:
[...]

PR #181, opened 2017-01-18
adding mingw deps to build.go for windows

codephobia:
Including the folder `/glfw/deps/mingw` to the build.go file for `dinput.h` issues during `go get` for glfw version 3.2. Fix for issue #179.

### DIFF
--- a/v3.2/glfw/build.go
+++ b/v3.2/glfw/build.go
@@ -4,7 +4,7 @@ package glfw
 // Windows Build Tags
 // ----------------
 // GLFW Options:
-#cgo windows CFLAGS: -D_GLFW_WIN32
+#cgo windows CFLAGS: -D_GLFW_WIN32 -Iglfw/deps/mingw
 
 // Linker Options:
 #cgo windows LDFLAGS: -lopengl32 -lgdi32


v3.2/glfw/build.go
```diff
@@ -4,7 +4,7 @@ package glfw
 // Windows Build Tags
 // ----------------
 // GLFW Options:
-#cgo windows CFLAGS: -D_GLFW_WIN32
+#cgo windows CFLAGS: -D_GLFW_WIN32 -Iglfw/deps/mingw
```
dmitshur, line 7: Given this is Windows, is it preferable to use `/` as path separator, or is it better to use backslash?

```
-Iglfw\deps\mingw
```

It seems that Windows can accept both.

I honestly don't know what's better. Any thoughts?

References:

- [os/path_windows.go](https://github.com/golang/go/blob/fcfd91858b7b57de0577e98703390f604971eaf4/src/os/path_windows.go#L8)
- https://en.wikipedia.org/wiki/Path_(computing)#MS-DOS.2FMicrosoft_Windows_style

tapir: `/` should be fine. I always use it that way and no problems so far

PR #187, opened 2017-03-29
README: Update OS X name to macOS.

dmitshur:
OS X has been renamed to macOS, as of version 10.12. Use its current name.

References:

-	https://www.apple.com/macos
-	https://www.businessinsider.com/wwdc-2016-os-x-becomes-macos-2016-6


tapir (approved):
Sorry it took so long 👍

PR #190, opened 2017-04-18
v3.2/glfw: Accept StandardCursor type in CreateStandardCursor.

### LINKED ISSUE #189: CreateStandardCursor(shape int) should be CreateStandardCursor(shape StandardCursor).
These consts have type `StandardCursor`:

```Go
const (
    ArrowCursor     StandardCursor = C.GLFW_ARROW_CURSOR
    IBeamCursor     StandardCursor = C.GLFW_IBEAM_CURSOR
    CrosshairCursor StandardCursor = C.GLFW_CROSSHAIR_CURSOR
    HandCursor      StandardCursor = C.GLFW_HAND_CURSOR
    HResizeCursor   StandardCursor = C.GLFW_HRESIZE_CURSOR
    VResizeCursor   StandardCursor = C.GLFW_VRESIZE_CURSOR
)
```

But [`glfw.CreateStandardCursor`](https://godoc.org/github.com/go-gl/glfw/v3.2/glfw#CreateStandardCursor) takes `int`, not `StandardCursor`.

Unless I'm missing something, it looks like its signature should be modified to take `StandardCursor` type.

Somewhat related to #170, but I think there's no conflict here.

Thanks to @dominikh for reporting this issue.

dmitshur:
The `StandardCursor` consts are used only for `CreateStandardCursor` method. There's no reason not to accept that type directly.

Fixes #189.

### DIFF
--- a/v3.2/glfw/input.go
+++ b/v3.2/glfw/input.go
@@ -425,7 +425,7 @@ func CreateCursor(img image.Image, xhot, yhot int) *Cursor {
 }
 
 // Returns a cursor with a standard shape, that can be set for a window with SetCursor.
-func CreateStandardCursor(shape int) *Cursor {
+func CreateStandardCursor(shape StandardCursor) *Cursor {
 	c := C.glfwCreateStandardCursor(C.int(shape))
 	panicError()
 	return &Cursor{c}


tapir:
This is a saddening oversight by my side but I'm afraid it will break code that did something like this ``` glfw.CreateStandardCursor(int(glfw.ArrowCursor))```

Should we still go on with it? I think we can.

dmitshur:
Yeah, sadly, this is a breaking API change.

However, it's going to be trivial for users to fix the code, and it will make their code simpler and cleaner. I think we should do it.

PR #194, opened 2017-05-01
v3.2/glfw: Fix Window.SetIcon handling of empty slice.

### LINKED ISSUE #193: `Window.SetIcon` panics with `index out of range` when passed a 0-length slice of icons
Hi!

There's a subtle bug with the Go version of [`Window.SetIcon` method](https://godoc.org/github.com/go-gl/glfw/v3.2/glfw#Window.SetIcon). The bug occurs when passed a 0-length slice. [This line](https://github.com/go-gl/glfw/blob/master/v3.2/glfw/window.go#L360) panics in such situation, because it attempts the get the address of the first element, which is out of range.

However, the doc says it should behave otherwise:
```
If no images are specified, the window reverts to its default icon.
```

dmitshur:
Window.SetIcon method is documented as:

	If no images are specified, the window reverts to its default icon.

Fix an index out of range panic when no images are provided. Use a null pointer in `C.glfwSetWindowIcon` call instead.

Fixes #193.

### DIFF
--- a/v3.2/glfw/window.go
+++ b/v3.2/glfw/window.go
@@ -357,7 +357,11 @@ func (w *Window) SetIcon(images []image.Image) {
 		cimages[i].pixels = (*C.uchar)(pix)
 	}
 
-	C.glfwSetWindowIcon(w.data, C.int(count), &cimages[0])
+	var p *C.GLFWimage
+	if count > 0 {
+		p = &cimages[0]
+	}
+	C.glfwSetWindowIcon(w.data, C.int(count), p)
 
 	for _, v := range freePixels {
 		v()


tapir (approved):
Didn't test but there is not much that can go wrong
Thanks

dmitshur:
I tested, didn't see issues.

However, I'm on macOS, and `SetIcon` doesn't seem to have any effect even when given non-empty slice of images, because there's no image in the window title bar (which is what `SetIcon` is for, I presume). According to http://www.glfw.org/docs/latest/group__window.html#gadd7ccd39fe7a7d1f0904666ae5932dc5, that's normal.

image

PR #220, opened 2018-08-12
v3.2/glfw: Create a glfw.Window from a *C.GLFWwindow reference

pchampio:
Bring back the glfw.Window. Opposite of #173


tapir:
Needs to accept `uintptr` and casting should be `(*C.GLFWwindow)(w)`. Both for the consistency reasons with the rest of the library

tapir:
Ah you're right. Go ahead and use unsafe.Pointer. I'll merge it

PR #226, opened 2018-10-08
X11: Fix IM-duplicated key events leaking through

### LINKED ISSUE #225: Is it OK to apply a change in GLFW 3.3 to GLFW 3.2?
Hi,

There is a relatively serious bug in GLFW 3.2 that key events on X11 are not fired correctly (https://github.com/glfw/glfw/issues/747, also discussed at https://github.com/hajimehoshi/ebiten/issues/648). This is fixed in GLFW 3.3 and we confirmed that the same patch (https://github.com/glfw/glfw/commit/4ff66a7818e3eaa5362c828a18220b12f1cd9bc8) works on GLFW 3.2. However, GLFW 3.3 is still beta and no one knows when it will be released.

My question is, would it be acceptable to apply that patch to go-gl's GLFW 3.2? I'm not sure the policy of go-gl GLFW and if we should keep this a simple wrapper of GLFW as it is, perhaps we should NOT apply the patch. If it is fine, I'd like to send the PR as soon as possible.

hajimehoshi:
This PR applies https://github.com/glfw/glfw/commit/4ff66a7818e3eaa5362c828a18220b12f1cd9bc8.

Fixes #225


tapir:
Thanks!

PR #229, opened 2018-10-12
add updateNSGLContext before poll events

### LINKED ISSUE #228: glfw doesn't draw any content initially on macOS Mojave
Due to the depreciation of OpenGL on macOS, a window created by glfw isn't drawn to until NSOpenGLContext is updated, seen [here](https://github.com/glfw/glfw/issues/1334#issuecomment-425298183). I have a workaround that calls the update [here](https://github.com/Noofbiz/glfwMojaveFix) that fixes the issue if it's called just before window.SwapBuffer(). I'm wondering if you want this integrated into this repo somewhere, and if so how would you like me to go about doing so? Thanks!

Noofbiz:
fixes #228

Tested on glfw 3.1 and 3.2 on both OSX High Sierra and OSX Mojave.

### DIFF
--- a/v3.1/glfw/context.go
+++ b/v3.1/glfw/context.go
@@ -37,6 +37,7 @@ func GetCurrentContext() *Window {
 // swap interval is greater than zero, the GPU driver waits the specified number
 // of screen updates before swapping the buffers.
 func (w *Window) SwapBuffers() {
+	w.updateNSGLContext()
 	C.glfwSwapBuffers(w.data)
 	panicError()
 }
--- a/v3.1/glfw/native_darwin.go
+++ b/v3.1/glfw/native_darwin.go
@@ -14,8 +14,15 @@ void *workaround_glfwGetCocoaWindow(GLFWwindow *w) {
 void *workaround_glfwGetNSGLContext(GLFWwindow *w) {
 	return (void *)glfwGetNSGLContext(w);
 }
+// workaround for inability to draw at start of program in OSX Mojave
+// See: https://github.com/glfw/glfw/issues/1334
+void cocoa_update_nsgl_context(void* id) {
+    NSOpenGLContext *ctx = id;
+    [ctx update];
+}
 */
 import "C"
+import "unsafe"
 
 func (w *Window) GetCocoaWindow() uintptr {
 	ret := uintptr(C.workaround_glfwGetCocoaWindow(w.data))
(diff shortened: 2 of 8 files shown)


kivutar (approved):
I tested it too. It works very well.

I was hoping it could also fix the vsync issue (https://github.com/glfw/glfw/issues/1337), but sadly it doesn't.

Another thing I've noticed: I'm getting a black screen when switching fullscreen/windowed. I made sure to call SwapBuffers on each frame. I can't explain this. This has been tested on a retina screen.

tapir:
I have lots of questions.

1-What does this code do actually? Why do we update only twice?
2-Also why is the numUpdates is global? Shouldn't it be window specific? If I have 2 windows created should they both increase the same counter?
3-I also don't like how windows and linux versions now need to call an empty func. Does Go compiler optimize it out?
4-If the blank screen @kivutar mentions is due to this patch that should also be fixed.
5-How about 3.3 branch?

Noofbiz:
I'll try my best to answer them.

1 - I got the fix from [here](https://github.com/glfw/glfw/issues/1334#issuecomment-425298183). In that comment they suggest updating the NSGLContext before calling PollEvents. The other solution is calling the window resize or move by a small amount then resizing/moving back.  This calls the update twice as well, so two times seems to be the magic number. The issue with resizing or moving is if we're using a full screen window we can't do that but this still works.

2 - I actually didn't think about multi-window applications, and haven't tested it. The original fix I did was in it's own package so the global variable didn't matter much. I agree, it should be window specific. I'll put it in there. Thanks for the catch!

3 - I don't like them calling empty functions either. One other way to do it would be to place PollEvents itself in the os-specific go files so they never have to call anything. Another would be to put it on the C side, which would be able to something like `#ifdef __APPLE__`. That said, the compiler should optimize it away.

4 - It fixes the blank screen but not the vsync issues.

5 - Haven't tried it on the 3.3 branch yet, but it should be. It's a problem with OSX depreciating open GL in general. This issue is showing up in sdl as well.

Noofbiz:
Honestly it's just a hot fix for something and would likely be removed soon when someone in glfw who actually knows the internals of cocoa gets something together. If you want to close this and just wait on glfw that would be fine.

tapir:
For this specific bug I would rather wait for the official fix since this one seem so hacky both on C and Go side. In the meantime you might want to look at the 3.3-beta branch maybe it works there?

PR #233, opened 2018-12-11
freebsd: Fix build

neilpa:
Use pkg-config to fix the FreeBSD build. This is a minimal change for now since I just wanted to get FreeBSD building. I'm tempted to do a follow-up that unifies the Linux options and uses pkg-config for all the library dependencies.

See also go-gl/glow#103 and go-gl/gl#117

### DIFF
--- a/v3.2/glfw/build.go
+++ b/v3.2/glfw/build.go
@@ -33,11 +33,13 @@ package glfw
 // FreeBSD Build Tags
 // ----------------
 // GLFW Options:
-#cgo freebsd,!wayland CFLAGS: -D_GLFW_X11 -D_GLFW_HAS_GLXGETPROCADDRESSARB -D_GLFW_HAS_DLOPEN
-#cgo freebsd,wayland CFLAGS: -D_GLFW_WAYLAND -D_GLFW_HAS_DLOPEN
+#cgo freebsd pkg-config: glfw3
+#cgo freebsd CFLAGS: -D_GLFW_HAS_DLOPEN
+#cgo freebsd,!wayland CFLAGS: -D_GLFW_X11 -D_GLFW_HAS_GLXGETPROCADDRESSARB
+#cgo freebsd,wayland CFLAGS: -D_GLFW_WAYLAND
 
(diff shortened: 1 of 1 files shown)


v3.2/glfw/build.go
```diff
@@ -33,11 +33,13 @@ package glfw
...
+#cgo freebsd pkg-config: glfw3
+#cgo freebsd CFLAGS: -D_GLFW_HAS_DLOPEN
+#cgo freebsd,!wayland CFLAGS: -D_GLFW_X11 -D_GLFW_HAS_GLXGETPROCADDRESSARB
+#cgo freebsd,wayland CFLAGS: -D_GLFW_WAYLAND
 
 // Linker Options:
-#cgo freebsd,!wayland LDFLAGS: -lGL -lX11 -lXrandr -lXxf86vm -lXi -lXcursor -lm -lXinerama
-#cgo freebsd,wayland LDFLAGS: -lGL -lwayland-client -lwayland-cursor -lwayland-egl -lxkbcommon -lm
+#cgo freebsd,!wayland LDFLAGS: -lm -lGL -lX11 -lXrandr -lXxf86vm -lXi -lXcursor -lXinerama
+#cgo freebsd,wayland LDFLAGS: -lm -lGL -lwayland-client -lwayland-cursor -lwayland-egl -lxkbcommon
```
hajimehoshi, line 43: I was wondering why you changed the order here.

neilpa: I was testing out a few different variants of this change to see what was actually required here. I kept it mostly as-is but made it clearer which libs are the same across both these targets.

hajimehoshi: ~I'm not sure (I became a member this month) but there might be an intention to keep this in ASCII order.~

Nvm, I don't see any intention in the current order, and I don't have a strong opinion on that.

tapir:
Oh wait. I was a little bit hasty to merge this one.
We include the GLFW source inside the package so why do we need to use pkg-config exactly?

PR #244, opened 2019-04-06
+Ability to link against OpenGL ES. Fixes #242.

### LINKED ISSUE #242: Cannot compile without GL.h
All versions of this wrapper cannot be compiled without `GL.h`. However, this header file does not necessarily exist on systems implementing OpenGL ES.

GLFW provides [header macros](https://www.glfw.org/docs/latest/build.html#build_macros) that should be defined for such cases (e.g. `GLFW_INCLUDE_ES1`). It should be possible to configure this wrapper accordingly. Possible solutions:

### Build Tags

I tried this in `build.go`:

```c
// OpenGL API Version Build Tags
// ----------------
#cgo opengles1 CFLAGS: -DGLFW_INCLUDE_ES1
#cgo opengles2 CFLAGS: -DGLFW_INCLUDE_ES2
#cgo opengles3 CFLAGS: -DGLFW_INCLUDE_ES3
#cgo opengles31 CFLAGS: -DGLFW_INCLUDE_ES31
#cgo openglcorearb CFLAGS: -DGLFW_INCLUDE_GLCOREARB
```

However, this will lead to GLFW itself being compiled with these macros, which it doesn't like that much:

```
In file included from ./c_glfw.go:4:
In file included from ./glfw/src/context.c:28:
./glfw/src/internal.h:45:3: error: "You must not define any header option macros when compiling GLFW"
```

This is because [cgo defines](https://golang.org/cmd/cgo/#hdr-Using_cgo_with_the_go_command):

> All the cgo CPPFLAGS and CFLAGS directives in a package are concatenated and used to compile C files in that package.

So to make this work, the build of the GLFW library must be moved to another package.

### GLFW_INCLUDE_NONE

The second alternative is to manually define this every time `glfw3.h` is included:

    //#define GLFW_INCLUDE_NONE
    //#include "glfw/include/GLFW/glfw3.h"

Problem with this is that `window.go` and `util.go` reference `C.GL_TRUE` which is only defined if *some* OpenGL implementation is included. However, I think that this reference should vanish anyway because GLFW supports Vulkan and that also does not have `GL_TRUE` afaik. I think it can be safely define as a `const` on the Go side.

---

I would happily create a PR if I know which option is preferred by y'all. Note that I am new to Go and may have overlooked something obvious.

flyx:
* Removed dependency on GL.h header
 * Added build tags to select OpenGL ES lib to link against
 * Updated README

See #242 for details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,18 @@
 go get -u github.com/go-gl/glfw/v3.2/glfw
 ```
 
+### OpenGL ES
+
+If your target system only provides an OpenGL ES implementation (true for some ARM boards), you need to link against that implementation.
+You do this by defining the appropriate build tags, e.g.
+
+```
+go get -u -tags=gles2 github.com/go-gl/glfw/v3.2/glfw
+```
+
+Supported tags are `gles1`, `gles2`, `gles3` and `vulkan`.
+Note that required packages might differ from those listed above; consult your hardware's documentation.
+
 ## Usage
 
 ```Go
--- a/v3.0/glfw/clipboard.go
+++ b/v3.0/glfw/clipboard.go
@@ -1,6 +1,7 @@
 package glfw
 
 //#include <stdlib.h>
+//#define GLFW_INCLUDE_NONE
 //#include <GLFW/glfw3.h>
 import "C"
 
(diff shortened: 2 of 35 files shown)


dmitshur:
CI wasn't passing on this due to a minor gofmt issue. I sent PR #247 to fix it.

PR #248, opened 2019-04-09
Add option to use external GLFW library

### LINKED ISSUE #246: Option to not use embedded GLFW source
Hello,

I'm currently in the process of packaging this project for Debian
and I wonder if there is a simple way to not compile
and then statically link the embedded GLFW source?

Is it possible to link system GLFW shared library dynamically?

dawidd6:
Closes: #246 

This commit adds an option to link against external GLFW library.
I think it would be convenient to also test this with CI.
I tested it only on linux, as I don't have access to machine with windows or darwin.

Simply appending `glfw_external` to `-tags` flag does the trick.

@tapir I didn't use `glfw-external` because it wasn't working, had to change `-` to `_`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,19 +29,10 @@ Note that required packages might differ from those listed above; consult your h
 
 If your host system has GLFW library already compiled and installed,
 you can link against it, instead of compiling and statically linking the embedded one.
-You do this by defining the appropriate build tag, e.g.
+You do this by passing `-tags=glfw_external` flag to `go build/install`,
+when building your application.
 
-For `v3.2`:
-```
-go get -u -tags=glfw_external github.com/go-gl/glfw/v3.2/glfw
-```
-
-For `v3.1`:
-```
-go get -u -tags=glfw_external github.com/go-gl/glfw/v3.1/glfw
-```
-
-This tag does not take effect with `v3.0` bindings.
+Supported for `v3.1` and later.
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Note that required packages might differ from those listed above; consult your h
 
 If your host system has GLFW library already compiled and installed,
 you can link against it, instead of compiling and statically linking the embedded one.
-You do this by passing `-tags=glfw_external` flag to `go build/install`,
+You do this by passing `-tags=glfw_external` flag to `go build` or `go install`,
 when building your application.
 
 Supported for `v3.1` and later.
--- a/README.md
+++ b/README.md
@@ -25,6 +25,24 @@ go get -u -tags=gles2 github.com/go-gl/glfw/v3.2/glfw
 Supported tags are `gles1`, `gles2`, `gles3` and `vulkan`.
 Note that required packages might differ from those listed above; consult your hardware's documentation.
 
+### External GLFW library
+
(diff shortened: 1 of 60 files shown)


README.md
```diff
@@ -25,6 +25,24 @@ go get -u -tags=gles2 github.com/go-gl/glfw/v3.2/glfw
...
+go get -u -tags=glfw_external github.com/go-gl/glfw/v3.2/glfw
+```
+
+For `v3.1`:
+```
+go get -u -tags=glfw_external github.com/go-gl/glfw/v3.1/glfw
+```
+
+This tag does not take effect with `v3.0` bindings.
+
```
pwaller, line 36: I'm unsure about this advice in the "go modules" and "go buildcache" world. I think in that case the instruction needs to be to build your application with this tag, not the library. I'm not sure that specifying a tag with `go get` for a library will really have the desired effect.

Are you familiar with Go modules and have you considered how packaging etc works in that situation? I don't have the Debian perspective so well internalized when it comes to packaging Go code.

[thinks to self, maybe I could search and see if I can find some quick advice. Aha: [Debian Go packaging](https://go-team.pages.debian.net/packaging.html)]

OK, so reading the above I see statements like this:

> Libraries written in Go are packaged for Debian with the only purpose of **building** other Go programs for Debian.

(emphasis from document)

I assume this is your goal too.

So to summarize my current thinking, which has evolved as I wrote this:

I do think it's worth documenting that you can build with `-tags=glfw_external`, but I think the advice needs to apply to the application, not the library. Then the advice doesn't need to be version specific except to say that it is only supported for v3.1 and later.

pwaller (dismissed):
Agree that it's worth testing this on CI, somehow!

README.md
dawidd6: Yes, you are absolutely right about everything you wrote.

Also, yes, I'm familiar with Go modules.
I'll make changes in README.

@pwaller Would you mind checking if it's properly documented now?

tapir:
I didn't have a chance to test this yet. In order to push it through I will have to test it on all platforms. So it might take a little while

README.md
pwaller: Your efforts are appreciated, thanks!

One more suggestion: `go build/install` looks like a command in its own right if you are new to go - better to say "`go build` or `go install`", perhaps. Otherwise :+1: from me on the docs front. Have not tested the code.

dawidd6: Done!

README.md
```diff
@@ -25,6 +25,15 @@ go get -u -tags=gles2 github.com/go-gl/glfw/v3.2/glfw
 Supported tags are `gles1`, `gles2`, `gles3` and `vulkan`.
 Note that required packages might differ from those listed above; consult your hardware's documentation.
 
+### External GLFW library
+
+If your host system has GLFW library already compiled and installed,
+you can link against it, instead of compiling and statically linking the embedded one.
+You do this by passing `-tags=glfw_external` flag to `go build` or `go install`,
```
dmitshur, line 32: The pattern for Go build constraints is to use no character to separate words. For example:

```
purego
netgo
nethttpomithttp2
```

We should follow that pattern by naming this `glfwexternal` rather than `glfw_external`.

PR #270, opened 2020-02-16
.travis.yml: use latest stable Go (1.13.8 and 1.12.17 at this time)

dmitshur:
Go 1.11.x and 1.10.x are very old by now; update to current versions.

Also resume testing v3.2/glfw package. It will be supported until it becomes too hard to support, then we'll say "please use v3.3 or newer".

As of Go 1.12, go tool vet is no longer supported.¹ Start using the go vet command instead.

Use RCS format for gofmt diff. It results in cleaner, easier to read gofmt diffs.

¹ https://golang.org/doc/go1.12#vet

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,17 +5,17 @@ addons:
     - xorg-dev
 language: go
 go:
-  - 1.11.x
-  - 1.10.x
-  - tip
+  - 1.x
+  - 1.12.x
+  - master
 matrix:
   allow_failures:
-    - go: tip
+    - go: master
   fast_finish: true
 install:
   - # Do nothing. This is needed to prevent default install action "go get -t -v ./..." from happening here (we want it to happen inside script step).
 script:
-  - go get -t -v ./v3.3/...
-  - diff -u <(echo -n) <(gofmt -d -s .)
-  - go tool vet .
-  - go test -v -race ./v3.3/...
+  - go get -t -v ./v3.3/... ./v3.2/...
+  - diff -n <(echo -n) <(gofmt -d -s .)
+  - go vet ./v3.3/... ./v3.2/...
+  - go test -v -race ./v3.3/... ./v3.2/...


hajimehoshi:
Why not using Go 1.13?

dmitshur:
It is testing with Go 1.13 (1.13.8 to be precise). "1.x" means latest stable Go version, which is currently Go 1.13.8. See https://docs.travis-ci.com/user/languages/go/#specifying-a-go-version-to-use.

You can also look at the CI build for this PR [here](https://travis-ci.org/go-gl/glfw/jobs/651250769?utm_medium=notification&utm_source=github_status), and notice it says "go version go1.13.8 linux/amd64".

In the [latest change](https://github.com/go-gl/glfw/compare/375e4b2f159c2a38c19fde5d5fca2d7b817dc4c8..0ae49ac6c441762e1304fcfa98920d86ea09320d), I went from expressing it as "1.13.x" to "1.x" for the following reason. Today, both map to Go 1.13.8. After some time passes and Go 1.14 becomes the latest release, I'd rather we test 1.14 than 1.13.

Ideally Travis CI would have a way of writing "1.x-1" to mean "the latest minor version of the previous stable major release", so we could target "1.x" and "1.x-1", and wouldn't have to keep updating the .travis.yml file every 6 months. But this is the best I came up with given what Travis offers.

hajimehoshi:
Acknowledged, I didn't know that. However, I'm still now sure why we adopted 1.x in this PR.

> Ideally Travis CI would have a way of writing "1.x-1" to mean "the latest minor version of the previous stable major release", so we could target "1.x" and "1.x-1", and wouldn't have to keep updating the .travis.yml file every 6 months. But this is the best I came up with given what Travis offers.

I'd prefer using 1.13 explicitly to mixing 1.x and 1.12.

dmitshur:
> I'm still now sure why we adopted 1.x in this PR.

I tried to explain in the 3rd paragraph of my last comment.

If we use history as an example, the last time we updated the Go versions in .travis.yml were in commit 547915429f423602ead3203dd3d1ca2ee3c4cffc on Oct 8, 2018. We went from testing 1.7 and 1.6.3 to 1.11.x and 1.10.x. Today, before this PR is merged, travis is still testing Go 1.11.x and 1.10.x, which are old and not supported by the Go team. If there's a problem in `glfw` that happens in Go 1.13 only, we wouldn't know.

So, that's why I prefer to use "1.x" over "1.13.x". Today they mean the same thing, and if we forget to update .travis.yml for many years (as we have done already!), it's better for it to be testing 1.15.x and 1.12.x (the latest supported version plus an old version) rather than 1.13.x and 1.12.x (both outdated versions).

Can you help me understand why you think writing "1.13.x" is better than "1.x"?

hajimehoshi:
> So, that's why I prefer to use "1.x" over "1.13.x". Today they mean the same thing, and if we forget to update .travis.yml for many years (as we have done already!), it's better for it to be testing 1.15.x and 1.12.x (the latest supported version plus an old version) rather than 1.13.x and 1.12.x (both outdated versions).
>
> Can you help me understand why you think writing "1.13.x" is better than "1.x"?

Now I understood the benefit of "1.x", but my concern is that "1.x" is not deterministic. If someone sent a PR, the test might fail due to the latest Go, not due to the PR. If we allow this indeterminism, using "1.x-1" would be more consistent. What do you think?

EDIT: Sorry, you meant "1.x-1" does not exist in TravisCI.

hajimehoshi:
We discussed offline and agreed that 1.x and 1.12 is the best solution to keep up with the latest Go so far: Updating .travis.yml for each Go project would be painful, but 1.x mitigates this issue.

.travis.yml
```diff
@@ -5,17 +5,17 @@ addons:
...
   - # Do nothing. This is needed to prevent default install action "go get -t -v ./..." from happening here (we want it to happen inside script step).
 script:
-  - go get -t -v ./v3.3/...
-  - diff -u <(echo -n) <(gofmt -d -s .)
-  - go tool vet .
-  - go test -v -race ./v3.3/...
+  - go get -t -v ./v3.3/... ./v3.2/...
+  - diff -n <(echo -n) <(gofmt -d -s .)
+  - go vet ./v3.3/... ./v3.2/...
+  - go test -v -race ./v3.3/... ./v3.2/...
```
pwaller, line 21: Hmm, is go test doing anything?

https://travis-ci.org/go-gl/glfw/jobs/651250770#L656-L658

```
$ go test -v -race ./v3.3/... ./v3.2/...
?   	github.com/go-gl/glfw/v3.3/glfw	[no test files]
?   	github.com/go-gl/glfw/v3.2/glfw	[no test files]
```

pwaller (approved):

.travis.yml
dmitshur: Yes, it is still testing that the package builds successfully.

We don't have any tests, hence "[no test files]". If we add some, they'll be run.

dmitshur:
Hmm, Travis was passing in this PR, but now that it's merged to master, it's failing there.

For some reason, it's running `travis_install_go_dependencies`, which it's not supposed to be doing because of this line:

```
install:
  - # Do nothing. This is needed to prevent default install action "go get -t -v ./..." from happening here (we want it to happen inside script step).
```

**Edit:** Pushed 6f7a984d4dc470c3f197229ad1991ae9e211bba2 to fix it.

PR #271, opened 2020-02-21
Add scripts/grab-upstream.sh

pwaller:
The purpose of this script is to make it self evident from Travis' status that
the files in the repository match upstream.

It also means that if you want to update the code, you just need to change
GLFW_C_REVISION.txt and run scripts/grab-upstream.sh.

Since it is checked, it will require less review to make such changes in the future.

Since it is automated, it will be easy for someone to update it if they need to
and not make mistakes regarding which files are included or not.

We also apply some patches to v3.2, and run the wayland generation.


pwaller:
Interesting, this catches some differences in v3.2!
```
	deleted:    v3.2/glfw/glfw/src/wayland-idle-inhibit-unstable-v1-client-protocol.c
	deleted:    v3.2/glfw/glfw/src/wayland-idle-inhibit-unstable-v1-client-protocol.h
	deleted:    v3.2/glfw/glfw/src/wayland-pointer-constraints-unstable-v1-client-protocol.c
	deleted:    v3.2/glfw/glfw/src/wayland-pointer-constraints-unstable-v1-client-protocol.h
	deleted:    v3.2/glfw/glfw/src/wayland-relative-pointer-unstable-v1-client-protocol.c
	deleted:    v3.2/glfw/glfw/src/wayland-relative-pointer-unstable-v1-client-protocol.h
	modified:   v3.2/glfw/glfw/src/wl_window.c
	modified:   v3.2/glfw/glfw/src/x11_platform.h
	modified:   v3.2/glfw/glfw/src/x11_window.c
```

pwaller:
There were a few fixes cherry-picked for v3.2. That might suggest we don't want to automate this, as it will make us follow upstream more rigidly. Those fixes have made it into v3.3 already, I take it?

If we can I would prefer if the fixes went upstream first. It seems sub-optimal for us to be maintaining a downstream fork.

pwaller:
Rebased for changes to travis.yml on master. PTAL @dmitshur @tapir

dmitshur:
I think this is good change overall, thanks for working on it.

I would like to suggest you consider a slightly different high level approach for doing the same task, and let me know if you think it'd be better or not.

We could use `//go:generate` directives to make it so that it's possible to run `go generate` on a glfw package to generate its `glfw` subdirectory with C sources.

That way, the definition of what it takes to generate the `glfw` subdirectory lives closer to the `glfw` package itself.

The Travis script can then do something as simple as `go generate ./...` and check if that causes any diffs.

What do you think about that?

pwaller:
Hi @dmitshur @tapir, I don't realistically have any time to play with this further at this point. Can we accept it as is (modulo conflicts), and if you like (and can find someone with time to do the work) it can be done later?

I ask because I hope that getting #303 done is a matter of updating `GLFW_C_REVISION.txt`and running this script.

dmitshur (approved):
That sounds reasonable to me, I think this is good to merge as long as it's working, and we can optionally improve it in the future.

v3.2/glfw/glfw/src/wl_window.c
```diff
@@ -24,6 +24,8 @@
 //
 //========================================================================
 
+#define _GNU_SOURCE
+
```
pwaller, line 28: Note that per https://github.com/go-gl/glfw/pull/283#issuecomment-723660957, this will introduce a double definition warning. From having looked into it before, it seems the best solution is to put `wl_window.c` into its own compilation unit and arrange that it isn't built with -D_GNU_SOURCE.

pwaller: Scratch that. Seperate compilation unit doesn't look straightforward. I've just `#undef _GNU_SOURCE` immediately before the `#include "wl_window.c"` instead.

pwaller:
Going to leave overnight for travis to run. If it passes, this can be merged as far as I'm concerned.

tapir:
Hi guys, I've been away for a while so please feel free to do whatever you think is best.

PR #280, opened 2020-04-04
fix: go mod vendor

### LINKED ISSUE #251: Not possible to import using `go mod vendor`

james-lawrence:
fixes #251 

add dummy files to ensure c source is maintained transparently.

why this works vs #258 because // +build ignore is treated specially by golangs toolchain and its necessary to import the packages for them to be maintained.

I'm only doing this for v3.3 because that is all I require for my use case atm. Someone else can pick up the other versions if/when necessary.

### DIFF
--- a/v3.3/glfw/build_cgo_hack.go
+++ b/v3.3/glfw/build_cgo_hack.go
@@ -10,7 +10,7 @@ package glfw
 //  - every directory we want to preserve is included here as a _ import.
 //  - this file is given a build to exclude it from the regular build.
 import (
-	// prevent go tooling from stripping out the c source files.
+	// Prevent go tooling from stripping out the c source files.
 	_ "github.com/go-gl/glfw/v3.3/glfw/glfw/deps"
 	_ "github.com/go-gl/glfw/v3.3/glfw/glfw/include/GLFW"
 	_ "github.com/go-gl/glfw/v3.3/glfw/glfw/src"
--- a/v3.3/glfw/build_cgo_hack.go
+++ b/v3.3/glfw/build_cgo_hack.go
@@ -1,0 +1,17 @@
+// +build required
+
+package glfw
+
+// this file exists purely to prevent the golang toolchain from stripping
+// away the c source directories and files.
+//
+// how it works:
+//  - every directory which only includes c source files receives a dummy.go file.
+//  - every directory we want to preserve is included here as a _ import.
+//  - this file is given a build to exclude it from the regular build.
+import (
+	// prevent go tooling from stripping out the c source files.
+	_ "github.com/go-gl/glfw/v3.3/glfw/glfw/deps"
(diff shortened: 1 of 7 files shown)


v3.3/glfw/build_cgo_hack.go
```diff
@@ -0,0 +1,17 @@
+// +build required
+
+package glfw
+
+// this file exists purely to prevent the golang toolchain from stripping
+// away the c source directories and files.
```
pwaller: Trivial: please can you document the circumstances under which the files are stripped? :)

```suggestion
// This file exists purely to prevent the golang toolchain from stripping
// away the c source directories and files when `go mod vendor` is used
// to populate a `vendor/` directory of a project depending on `go-gl/glfw`.
```

v3.3/glfw/build_cgo_hack.go
```diff
@@ -0,0 +1,17 @@
...
+//
+// how it works:
```
pwaller:
```suggestion
// How it works:
```

v3.3/glfw/build_cgo_hack.go
```diff
@@ -0,0 +1,17 @@
...
+//  - every directory which only includes c source files receives a dummy.go file.
+//  - every directory we want to preserve is included here as a _ import.
+//  - this file is given a build to exclude it from the regular build.
+import (
+	// prevent go tooling from stripping out the c source files.
```
pwaller:
```suggestion
	// Prevent go tooling from stripping out the c source files.
```

v3.3/glfw/glfw/deps/dummy.go
```diff
@@ -0,0 +1,11 @@
+// +build required
+
+// Package dummy prevents go tooling from stripping the c dependencies.
+package dummy
+
+import (
+	// prevent go tooling from stripping out the c source files.
```
pwaller:
```suggestion
	// Prevent go tooling from stripping out the c source files.
```

pwaller (approved):
Nice fix, thanks! Some trivial suggestions to the comments.

james-lawrence:
accepted your suggestions.

pwaller:
Thanks!

Issue #278, opened 2020-03-29
Cannot compile app with `go run -tags wayland`


andydotxyz:
I cannot compile for wayland, error is:

```
# github.com/go-gl/glfw/v3.3/glfw
In file included from ./glfw/src/internal.h:190,
                 from ./glfw/src/context.c:30,
                 from ../../../../../pkg/mod/github.com/go-gl/glfw/v3.3/glfw@v0.0.0-20200222043503-6f7a984d4dc4/c_glfw.go:4:
./glfw/src/wl_platform.h:59:10: fatal error: wayland-xdg-shell-client-protocol.h: No such file or directory
   59 | #include "wayland-xdg-shell-client-protocol.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

I have tried running the wayland generate script, however that only creates the files for wayland-idle-inhibit-, wayland-pointer-constraints and wayland-relative-pointer. I cannot figure out how the additional files referenced by this project are created or found?

Compiling against v3.3.1 - 0a49ef0a00baa3ab520ddc452f0e3b1e099c5589

andydotxyz:
Same result without modules enabled btw

pwaller:
Looks like they are generated here if you'd like to take a crack at fixing them. Just add a few more, I think:

https://github.com/go-gl/glfw/blob/9b86c984c866400dc34182d3a6aaf4cb5be0a4cf/scripts/generate-wayland-protocols.sh#L18-L19

And the ones you need:

https://github.com/go-gl/glfw/blob/9b86c984c866400dc34182d3a6aaf4cb5be0a4cf/v3.3/glfw/glfw/src/wl_platform.h#L59-L64

I'd be very grateful if you could test and whip up a pull request.

andydotxyz:
Thanks. The PR #283 fixes the include error - but introduces a *lot* of compile errors, all similar to the first one, which is:

```
# github.com/go-gl/glfw/v3.3/glfw
In file included from ../../../../github.com/go-gl/glfw/v3.3/glfw/c_glfw_linbsd.go:9:
../../../../github.com/go-gl/glfw/v3.3/glfw/glfw/src/wl_window.c:29: warning: "_GNU_SOURCE" redefined
   29 | #define _GNU_SOURCE
      | 
<command-line>: note: this is the location of the previous definition
[...]

PR #283, opened 2020-04-29
Fix wayland builds

andydotxyz:
This picks all the files out of the wayland source that should compile with `-tags wayland`


andydotxyz:
This fixes the errors in #278 - Though I now have new build errors (see the ticket)

pwaller:
@andydotxyz thanks for doing the legwork.

One quick request: could you rebase your changes rather than introducing a merge?

Other than that - the only other thing I can think of it would be great to have is confirmation from someone else that it works for them. If you can provide me with a method of testing which would show that this works, then I'll do that to test it when I get a moment. Then we can merge. Patch looks reasonable on the face of it.

Please keep pinging this thread if you don't hear back in the near future (once a week is a good benchmark).

danfragoso:
I was having the same problem when building with `-tags wayland`. I can confirm that @andydotxyz patch fixed the issue.

andydotxyz:
Apologies I had to force push a couple of times as the conflicts kept kicking me.
This is now based on current master with no merges on the PR history.

andydotxyz:
With regards to testing... using any Go-GLFW app `go run -tags wayland .` - this didn't compile or run before this patch, but once applied it will :).

scripts/generate-wayland-protocols.sh
```diff
@@ -11,18 +11,29 @@ fi
...
-
-wayland-scanner code $TMP_CLONE_DIR/unstable/relative-pointer/relative-pointer-unstable-v1.xml "$GLGLFW_PATH"/wayland-relative-pointer-unstable-v1-client-protocol.c
-wayland-scanner client-header $TMP_CLONE_DIR/unstable/relative-pointer/relative-pointer-unstable-v1.xml "$GLGLFW_PATH"/wayland-relative-pointer-unstable-v1-client-protocol.h
-
-wayland-scanner code $TMP_CLONE_DIR/unstable/idle-inhibit/idle-inhibit-unstable-v1.xml "$GLGLFW_PATH"/wayland-idle-inhibit-unstable-v1-client-protocol.c
-wayland-scanner client-header $TMP_CLONE_DIR/unstable/idle-inhibit/idle-inhibit-unstable-v1.xml "$GLGLFW_PATH"/wayland-idle-inhibit-unstable-v1-client-protocol.h
+generate() {
+  HEADER=$1
+  VER=$2
+  if [[ "$VER" == "stable" ]]; then
```
pwaller: Minor issue here: On Ubuntu 20.04, /bin/sh is dash, so this fails because [[ is a bashism :(.

pwaller:
Confirmed that it builds for me. Also confirmed that I get the same generated sources when I regenerate them. Minor issue introduced into generate-wayland-protocols.sh. I'm unable to test functionality but I trust that this is an improvement on that front.

I do see a warning:
```
In file included from ./c_glfw_lin.go:9:
./glfw/src/wl_window.c:29: warning: "_GNU_SOURCE" redefined
   29 | #define _GNU_SOURCE
      | 
```

It seems the issue is that we supply `-D_GNU_SOURCE` and additionally `wl_window.c` from upstream defines it.

Unfortunately if I get rid of the double define, the first `#include` of the internal headers in an earlier source file results in the wrong symbols being defined by the time wl_window. So either:
1. we should move wl_window to its own compilation unit, 
2. we should #undef and #redef _GNU_SOURCE on either side of `#include "wl_window.c".

A bit grim. But this is non-fatal and the situation is generally improved with this PR, so I think we should accept it at this point once generate-wayland-protocols.sh is fixed. Patches welcome to remove the warning also but not required.

andydotxyz:
The origincal script did not work in dash either, due to the "==". But now it does.

andydotxyz:
The _GNU_SOURCE issue seems to reate to https://github.com/go-gl/glfw/commit/bac7f0e2e1cc1b17dfc90d2c83c04f174e53acf8 
and I don't have a good solution sorry. I'd be happy for this to merge now as you say it can be improved later - at least it compile now!

pwaller:
Yes, though that actually hasn't been merged (note the message saying the commit doesn't appear on any branch in this repo). That was me in #271 making a script to automatically fetch from the sources, and finding that someone had previously deleted it from the upstream sources (in v3.2). It still stands in v3.3 afaict.

And here it is upstream: https://github.com/glfw/glfw/blob/0ef149c8f2451fbc8e866834675a075cfc295b6c/src/wl_window.c#L29

pwaller:
Thanks for your patience and following through with this :)

PR #307, opened 2021-03-06
Use C.CBytes() to implement bytes(), fix vet warning

pwaller:
CI is currently failing with the following on go1.16:

```
v3.2/glfw/util.go:29:41: possible misuse of reflect.SliceHeader
v3.3/glfw/util.go:29:41: possible misuse of reflect.SliceHeader
```

I don't know whether this is a false positive or not but we should try to be go-vet clean.

It looks to me like `bytes()` could be implemented with this instead, avoiding the issue altogether:

```
func C.CBytes([]byte) unsafe.Pointer
```

At this point, I've only built, but not tested this behaviour extensively, to get CI passing. It looks low risk but don't know if there is anything subtly different about the original implementation. @tapir any thoughts?

### DIFF
--- a/v3.2/glfw/util.go
+++ b/v3.2/glfw/util.go
@@ -5,11 +5,6 @@ package glfw
 //#include "glfw/include/GLFW/glfw3.h"
 import "C"
 
-import (
-	"reflect"
-	"unsafe"
-)
-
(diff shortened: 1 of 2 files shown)


pwaller:
Going to merge this untested, please chime in if you spot anything amiss.

PR #310, opened 2021-03-06
Make build cache a function of the glfw C sources

### LINKED ISSUE #269: make GLFW C revision a part of input to Go build cache
From https://golang.org/cmd/go/#hdr-Build_and_test_caching:

> The build cache correctly accounts for changes to Go source files, compilers, compiler options, and so on: cleaning the cache explicitly should not be necessary in typical use. However, the build cache does not detect changes to C libraries imported with cgo. If you have made changes to the C libraries on your system, you will need to clean the cache explicitly or else use the -a build flag (see 'go help build') to force rebuilding of packages that depend on the updated C libraries.

There's currently a problem where updating the GLFW C version does not invalidate the Go build cache for the `github.com/go-gl/glfw/v3.3/glfw` package.

We have a `GLFW_C_REVISION.txt` file that tracks the revision of the GLFW C library being used.

We can rename the `GLFW_C_REVISION.txt` file to something like `glfwcrev.go` and include the GLFW C revision as a comment:

```
package glfw

// GLFW_C_REVISION = 0a49ef0a00baa3ab520ddc452f0e3b1e099c5580
```

Then changing the GLFW C revision will cause one of the .go files to become different, and the old build cache will no longer get in the way.

This is somewhat related to #251 and #258.

pwaller:
This is a workaround for the issue described in #269. Go unfortunately doesn't
consider the C source files as part of its build cache calculation. Therefore,
invalidate it by writing a hash which is computed across the full C source
directory.

This hash is only updated when `go generate` is called, but this will at least
help the majority of users, when glfw is updated. This is encoded in
grab-upstream.sh which is used for updating the sources.

This hash is intended to be protected from divergance by CI, too.

Resolves #269.


v3.3/glfw/glfw_tree_rebuild.go
```diff
@@ -0,0 +1,10 @@
+package glfw
+
+//go:generate ../../scripts/glfw_tree_rebuild.sh
+
+// This is a recursive hash of the full contents of the upstream glfw, as
```
hajimehoshi, line 5: nits: `upstreamTreeSHA is ...` for Go's comment convention?

pwaller: Ah, my go idioms are a little rusty. Fixed 211ecac01a39078a11f649cfd964e30d2ca90bd9. Thanks ;-)

Issue #292, opened 2020-09-05
Uncaught error when connecting Bluetooth controllers


shavs:
In [Ludo](https://github.com/libretro/ludo), there's a breaking exception that's thrown from `glfw` when connecting a Bluetooth controller:

```
GLFW: An uncaught error has occurred: InvalidValue: Invalid button in gamepad mapping 050000005e040000fd02000003090000 (Xbox One Wireless Controller)
GLFW: Please report this bug in the Go package immediately.
GLFW: An uncaught error has occurred: InvalidValue: Invalid button in gamepad mapping 050000005e040000fd02000003090000 (Xbox One Wireless Controller)
GLFW: Please report this bug in the Go package immediately.
GLFW: An uncaught error has occurred: InvalidValue: Invalid button in gamepad mapping 050000005e040000fd02000003090000 (Xbox One Wireless Controller)
GLFW: Please report this bug in the Go package immediately.
GLFW: An uncaught error has occurred: InvalidValue: Invalid button in gamepad mapping 050000005e040000fd02000003090000 (Xbox One Wireless Controller)
GLFW: Please report this bug in the Go package immediately.
GLFW: An uncaught error has occurred: InvalidValue: Invalid button in gamepad mapping 050000005e040000fd02000003090000 (Xbox One Wireless Controller)
GLFW: Please report this bug in the Go package immediately.
GLFW: An uncaught error has occurred: InvalidValue: Invalid button in gamepad mapping 050000005e040000fd02000003090000 (Xbox One Wireless Controller)
GLFW: Please report this bug in the Go package immediately.
GLFW: An uncaught error has occurred: InvalidValue: Invalid button in gamepad mapping 050000005e040000fd02000003090000 (Xbox One Wireless Controller)
GLFW: Please report this bug in the Go package immediately.
GLFW: An uncaught error has occurred: InvalidValue: Invalid button in gamepad mapping 050000005e040000fd02000003090000 (Xbox One Wireless Controller)
GLFW: Please report this bug in the Go package immediately.
GLFW: An uncaught error has occurred: InvalidValue: Invalid button in gamepad mapping 050000005e040000fd02000003090000 (Xbox One Wireless Controller)
GLFW: Please report this bug in the Go package immediately.
GLFW: An uncaught error has occurred: InvalidValue: Invalid button in gamepad mapping 050000005e040000fd02000003090000 (Xbox One Wireless Controller)
GLFW: Please report this bug in the Go package immediately.
[...]

PR #325, opened 2021-07-14
Allow GLFW_INVALID_VALUE at Init

### LINKED ISSUE #324: Can we apply the gamepad issue fix before 3.3.5 is released?
Now GLFW 3.3 has an issue that specific gamepads crashes the app immediately, and this will be fixed in 3.3.5 (https://github.com/glfw/glfw/issues/1763). This is a pretty serious problem in my library that uses GLFW (e.g. https://github.com/hajimehoshi/ebiten/issues/1229). As we're not sure when 3.3.5 is released, would it be possible to apply the patch before 3.3.5 is released?

CC @dmitshur

hajimehoshi:
Closes #324
Updates #292

### DIFF
--- a/v3.3/glfw/glfw.go
+++ b/v3.3/glfw/glfw.go
@@ -34,7 +34,10 @@ const (
 // This function may only be called from the main thread.
 func Init() error {
 	C.glfwInit()
-	return acceptError(APIUnavailable)
+	// invalidValue can happen when specific joysticks are used. This issue
+	// will be fixed in GLFW 3.3.5. As a temporary fix, accept this error.
+	// See go-gl/glfw#292, go-gl/glfw#324, and glfw/glfw#1763.
+	return acceptError(APIUnavailable, invalidValue)
 }
 
 // Terminate destroys all remaining windows, frees any allocated resources and


hajimehoshi:
PTAL

dmitshur (approved):
LGTM. There's a typo in the commit message/PR subject, s/GLFW_INVLAID_VALUE /GLFW_INVALID_VALUE/.

dmitshur:
Thinking about this more (sorry I missed it earlier), this workaround should probably not return `invalidValue` from `Init`, but rather handle it internally.

That is, most callers of `glfw.Init` look like this:

```Go
err := glfw.Init()
if err != nil {
    log.Fatalln("glfw.Init failed:", err)
}
```

This temporary workaround shouldn't require users to also temporarily update their code.

Perhaps something like:

```Go
err := acceptError(APIUnavailable, invalidValue)
if e, ok := err.(*Error); ok && e.Code == invalidValue {
    // This is an internal and non-fatal error that's being intentionally ignored.
    err = nil
}
return err
```

(It's also possible to handle it by logging, but given GLFW 3.3.5 outright removes it, logging doesn't seem to be needed.)

Issue #313, opened 2021-03-28
v3.3/glfw: binaries built for FreeBSD links against shared libglfw.so.3


Jacalz:
During some FreeBSD testing of [wormhole-gui](https://github.com/Jacalz/wormhole-gui) (uses this project through [Fyne](https://github.com/fyne-io/fyne)), I noted that running FreeBSD binaries (both those compiled from Linux and those compiled on FreeBSD) linked dynamically against a shared `libglfw.so.3`. This means that running the binaries without first installing glfw with `pkg install glfw` resulted in the following error:
```
ld-elf.so.1: Shared object "libglfw.so.3" not found, required by "wormhole-gui"
```

This was quite unexpected given that every other platform seems to be linking statically against it and does not require installing glfw in any way to run the binaries. I thus opened a bug for this for Fyne (see https://github.com/fyne-io/fyne/issues/1928), but I don't know how much we can do to fix this, so I decided to open the issue here as well to inform you guys.

pwaller:
Thanks for the note. I agree it's weird.

Seems this was a mistake introduced in https://github.com/go-gl/glfw/pull/233. Happy to review PRs to change the situation but don't have a freebsd system to test on, so help wanted to fix this.

Ping @neilpa who made the change and @hajimehoshi @tapir who also looked at it. Other than making the build work I guess there was not much rational behind using pkg-config?

Jacalz:
I tried to get this working when cross-compiling to FreeBSD on Linux but couldn't quite figure out what was wrong. Reverting the mentioned commit above just leads to the following error (which is weird, `-lX11` is already in the linker flags):
```
# github.com/go-gl/glfw/v3.3/glfw
In file included from vendor/github.com/go-gl/glfw/v3.3/glfw/c_glfw.go:4:
In file included from ./glfw/src/context.c:30:
In file included from ./glfw/src/internal.h:188:
./glfw/src/x11_platform.h:33:10: fatal error: 'X11/Xlib.h' file not found
#include <X11/Xlib.h>
         ^~~~~~~~~~~~
```
I think someone more familiar with FreeBSD and the glfw build flags need to have a look at this.

xorander00:
@Jacalz I just ran into this error as well. From a quick, initial look at the error message you just posted above, it seems like /usr/local isn't being included in the search paths for CC. If that's the case, then it's likely also the case for the linker path.  
  
[...]

PR #336, opened 2021-10-21
v3.3/glfw: add openbsd support

### LINKED ISSUE #288: Can go-gl/glfw support other UNIX systems like NetBSD?
GLFW itself supports UNIX-like systems in theory, but go-gl/glfw supports only Linux and FreeBSD. Just out of curiosity, I was wondering why there is such limitation.

ilylily:
partially fixes #288

simple adaptations to add openbsd support. this makes `fyne.io/fyne/v2/cmd/fyne_demo/` run on my OpenBSD-current machine. i don't have a netbsd system available, but hopefully this makes it clear which 8-ish changes a netbsd user would need to make to support that too

### DIFF
--- a/v3.3/glfw/build.go
+++ b/v3.3/glfw/build.go
@@ -41,15 +41,15 @@ package glfw
 // BSD Build Tags
 // ----------------
 // GLFW Options:
+#cgo openbsd pkg-config: x11 xau xcb xdmcp
 #cgo freebsd pkg-config: glfw3
 #cgo freebsd openbsd CFLAGS: -D_GLFW_HAS_DLOPEN
 #cgo freebsd,!wayland openbsd CFLAGS: -D_GLFW_X11 -D_GLFW_HAS_GLXGETPROCADDRESSARB
 #cgo freebsd,wayland CFLAGS: -D_GLFW_WAYLAND
-#cgo openbsd CFLAGS: -I/usr/X11R6/include
 
 // Linker Options:
 #cgo freebsd,!wayland LDFLAGS: -lm -lGL -lX11 -lXrandr -lXxf86vm -lXi -lXcursor -lXinerama
 #cgo freebsd,wayland LDFLAGS: -lm -lGL -lwayland-client -lwayland-cursor -lwayland-egl -lxkbcommon
-#cgo openbsd LDFLAGS: -L/usr/X11R6/lib -lm -lX11 -lxcb -lXau -lXdmcp
+#cgo openbsd LDFLAGS: -lm
 */
 import "C"
--- a/v3.3/glfw/build.go
+++ b/v3.3/glfw/build.go
@@ -41,8 +41,8 @@ package glfw
 // BSD Build Tags
 // ----------------
 // GLFW Options:
-#cgo openbsd pkg-config: x11 xau xcb xdmcp
 #cgo freebsd pkg-config: glfw3
+#cgo openbsd pkg-config: x11 xau xcb xdmcp
 #cgo freebsd openbsd CFLAGS: -D_GLFW_HAS_DLOPEN
 #cgo freebsd,!wayland openbsd CFLAGS: -D_GLFW_X11 -D_GLFW_HAS_GLXGETPROCADDRESSARB
 #cgo freebsd,wayland CFLAGS: -D_GLFW_WAYLAND
--- a/v3.3/glfw/build.go
+++ b/v3.3/glfw/build.go
@@ -45,7 +45,6 @@ package glfw
 #cgo freebsd openbsd CFLAGS: -D_GLFW_HAS_DLOPEN
 #cgo freebsd,!wayland openbsd CFLAGS: -D_GLFW_X11 -D_GLFW_HAS_GLXGETPROCADDRESSARB
 #cgo freebsd,wayland CFLAGS: -D_GLFW_WAYLAND
-#cgo openbsd CFLAGS: -I/usr/X11R6/include
 
 // Linker Options:
 #cgo freebsd,!wayland LDFLAGS: -lm -lGL -lX11 -lXrandr -lXxf86vm -lXi -lXcursor -lXinerama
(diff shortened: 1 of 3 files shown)


hajimehoshi:
Nice! This looks good to me but I'd like to defer this to other reviewers.

Can we support NetBSD in the same way?

ilylily:
netbsd support seems blocked by missing support in [go-gl/glow](https://github.com/go-gl/glow) and [godbus/dbus](https://github.com/godbus/dbus). but yes, a very similar set of changes will work for netbsd once those support it

hajimehoshi:
If there is no objection, I'll merge this.

v3.3/glfw/build.go
```diff
@@ -38,16 +38,18 @@ package glfw
...
 // ----------------
 // GLFW Options:
-#cgo freebsd pkg-config: glfw3
-#cgo freebsd CFLAGS: -D_GLFW_HAS_DLOPEN
-#cgo freebsd,!wayland CFLAGS: -D_GLFW_X11 -D_GLFW_HAS_GLXGETPROCADDRESSARB
+#cgo freebsd openbsd pkg-config: glfw3
+#cgo freebsd openbsd CFLAGS: -D_GLFW_HAS_DLOPEN
+#cgo freebsd,!wayland openbsd CFLAGS: -D_GLFW_X11 -D_GLFW_HAS_GLXGETPROCADDRESSARB
 #cgo freebsd,wayland CFLAGS: -D_GLFW_WAYLAND
+#cgo openbsd CFLAGS: -I/usr/X11R6/include
```
dmitshur: Is "/usr/X11R6/include" considered a standard include directory on OpenBSD?

dmitshur (approved):
Looks reasonable to me. One minor comment/question.

v3.3/glfw/build.go
ilylily: yes, that is the standard installation path for openbsd's x11 fork, which is included by default. similarly, netbsd uses `/usr/X11R7`

ilylily: actually just realized that line is unnecessary; the pkg-config covers it

v3.3/glfw/build.go
```diff
@@ -38,16 +38,17 @@ package glfw
...
-#cgo freebsd,!wayland CFLAGS: -D_GLFW_X11 -D_GLFW_HAS_GLXGETPROCADDRESSARB
+#cgo freebsd openbsd pkg-config: glfw3
+#cgo freebsd openbsd CFLAGS: -D_GLFW_HAS_DLOPEN
+#cgo freebsd,!wayland openbsd CFLAGS: -D_GLFW_X11 -D_GLFW_HAS_GLXGETPROCADDRESSARB
 #cgo freebsd,wayland CFLAGS: -D_GLFW_WAYLAND
 
 // Linker Options:
 #cgo freebsd,!wayland LDFLAGS: -lm -lGL -lX11 -lXrandr -lXxf86vm -lXi -lXcursor -lXinerama
 #cgo freebsd,wayland LDFLAGS: -lm -lGL -lwayland-client -lwayland-cursor -lwayland-egl -lxkbcommon
+#cgo openbsd LDFLAGS: -L/usr/X11R6/lib -lm -lX11 -lxcb -lXau -lXdmcp
```
hajimehoshi: Is `-L` needed, or is this included in pkg-config?

ilylily: that's actually why i thought the `-I` earlier was needed. `pkg-config --libs glfw3` only returns `-L/usr/local/lib -lglfw`, hence the need for these LDFLAGS, which are pulled from `pkg-config --cflags --libs glfw3`. which returns different output, for some reason: `-I/usr/local/include -I/usr/X11R6/include -L/usr/local/lib -L/usr/X11R6/lib -lglfw -lX11 -lxcb -lXau -lXdmcp`

ilylily: tl;dr yes, pkg-config is weird

hajimehoshi:
I realized that `pkg-config` for `glfw3` seems weird (#313). We don't have to need this for *BSD. I'd like to wait for #313 solved.

Jacalz (changes requested):
Really nice work in this. I'm very interested in giving it a go running Fyne on OpenBSD :)

Can we please get the install instructions in the README updated for OpenBSD (possibly NetBSD also if that starts working)? I realise that FreeBSD isn't there but that might be outside the work of this PR, I guess?

hajimehoshi:
@ilylily would it be possible to have this change with `pkg-config` not for `glfw3` but for necessary dependencies? Depending external `glfw3` seems weird. (Then we have to address this for FreeBSD separately)

v3.3/glfw/build.go
hajimehoshi: Now we have `-I/usr/X11R6/include` again but can we remove this by `pkg-config` for X? Hopefully could we remove `-L` in the same way?

hajimehoshi:
Thanks for the fix. I left a comment.

v3.3/glfw/build.go
```diff
@@ -38,16 +38,18 @@ package glfw
 #cgo linux,!wayland LDFLAGS: -lX11 -lXrandr -lXxf86vm -lXi -lXcursor -lm -lXinerama -ldl -lrt
 #cgo linux,wayland LDFLAGS: -lwayland-client -lwayland-cursor -lwayland-egl -lxkbcommon -lm -ldl -lrt
 
-// FreeBSD Build Tags
+// BSD Build Tags
 // ----------------
 // GLFW Options:
+#cgo openbsd pkg-config: x11 xau xcb xdmcp
 #cgo freebsd pkg-config: glfw3
```
hajimehoshi, line 44: One last request: can we swap the lines `openbsd pkg-config` and `freebsd pkg-config`? (probably we should keep alphabetical order here)

v3.3/glfw/build.go
hajimehoshi: It looks like you have already fixed. Thanks!

v3.3/glfw/build.go
ilylily: got it

hajimehoshi (approved):
LGTM, thanks!

hajimehoshi:
Note that we should address #313 later. I hope we can use the same way as OpenBSD.

Jacalz:
As I said earlier, I also think we should update the README with installation instructions for OpenBSD.

hajimehoshi:
> As I said earlier, I also think we should update the README with installation instructions for OpenBSD.

Oh sorry I missed that.

@ilylily Could you create a PR to update README? Thanks,

PR #337, opened 2021-10-23
include openbsd build info

ilylily:
mentioned in #336

cc @Jacalz @hajimehoshi

feels kind of bare, but i just tested in a clean vm and no external dependencies are needed. except for Go, of course, but that isn't mentioned on any of the other platforms

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@
 	* On macOS, you need Xcode or Command Line Tools for Xcode (`xcode-select --install`) for required headers and libraries.
 	* On Ubuntu/Debian-like Linux distributions, you need `libgl1-mesa-dev` and `xorg-dev` packages.
 	* On CentOS/Fedora-like Linux distributions, you need `libX11-devel libXcursor-devel libXrandr-devel libXinerama-devel mesa-libGL-devel libXi-devel libXxf86vm-devel` packages.
+        * On OpenBSD, you need the X11 sets. These are installed by default, and can be added from the ramdisk kernel at any time.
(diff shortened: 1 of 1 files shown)


hajimehoshi (approved):
LGTM, thanks! I'll merge this if @Jacalz is fine.

README.md
```diff
@@ -6,6 +6,7 @@
 	* On macOS, you need Xcode or Command Line Tools for Xcode (`xcode-select --install`) for required headers and libraries.
 	* On Ubuntu/Debian-like Linux distributions, you need `libgl1-mesa-dev` and `xorg-dev` packages.
 	* On CentOS/Fedora-like Linux distributions, you need `libX11-devel libXcursor-devel libXrandr-devel libXinerama-devel mesa-libGL-devel libXi-devel libXxf86vm-devel` packages.
+        * On OpenBSD, you need the X11 sets. These are installed by default, and can be added from the ramdisk kernel at any time.
```
hajimehoshi: One question: I'm not sure 'the default'. X is not installed in the headless mode, right?

ilylily: from the [faq](https://www.openbsd.org/faq/faq4.html):
>  Some libraries from xbase70.tgz, like freetype or fontconfig, can be used outside of X by programs that manipulate text or graphics. Such programs will usually need fonts, either from xfont70.tgz or font packages. For the sake of simplicity, the developers decided against maintaining a minimal xbase70.tgz set that would allow most non-X ports to run.

they're selected by default on all installs and must be explicitly disabled

hajimehoshi: OK thanks!

Jacalz (approved):
Splendid information. Thanks :)

PR #338, opened 2021-10-24
v3.3/glfw: remove external glfw dependency on freebsd

### LINKED ISSUE #313: v3.3/glfw: binaries built for FreeBSD links against shared libglfw.so.3

ilylily:
fixes #313 and https://github.com/fyne-io/fyne/issues/1928


hajimehoshi (approved):
LGTM, thanks!

Jacalz:
Big thanks @ilylily for fixing this :)

PR #341, opened 2021-11-28
v3.3/glfw: add netbsd support

ilylily:
just add some missing build tags. helps with #288

should wayland build tags be included even if glfw upstream doesn't support netbsd's input system yet? and if so, should i keep the mention in the readme? either glfw supporting wscons or netbsd supporting evdev would make this build for wayland as is

as an example of updating a wayland application in such a way, Nia Alarie added wscons to swc with [a surprisingly small patchset](https://github.com/michaelforney/swc/commit/374e92246a19df53d4882d4ece39f31b4aabccee). so i think this would be easier than it might sound at first

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,7 @@
 	* On Ubuntu/Debian-like Linux distributions, you need `libgl1-mesa-dev` and `xorg-dev` packages.
 	* On CentOS/Fedora-like Linux distributions, you need `libX11-devel libXcursor-devel libXrandr-devel libXinerama-devel mesa-libGL-devel libXi-devel libXxf86vm-devel` packages.
 	* On FreeBSD, you need the package `pkgconf`. To build for X, you also need the package `xorg`; and to build for Wayland, you need the package `wayland`.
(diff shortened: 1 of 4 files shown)


Jacalz (approved):
This looks good to me. Thanks a lot for making it happen.
I think it would be sensible to leave waylaid support out given that it isn't supported upstream. Is there an issue upstream about adding that support?

dmitshur (approved):
Thanks, this change looks reasonable, so I'll merge it to move it forward.

I think it's fine both to leave out `wayland` tags until they're supported, but also okay to include them proactively given the README says the support is incomplete. Having tags in place in advance is maybe less than ideal, but not a big deal. Both options are okay, so let's proceed with it as is for now to save on a review roundtrip, and adjust later as needed.

Jacalz:
I guess this also requires work over in the `gl` package?
image

ilylily:
`glow` is patched with netbsd support, so `gl` just needs to be regenerated. you can mess with the instructions in the `glow` readme to generate the relevant profiles, and vendor them in for testing. not sure who to talk to about updating `gl`

Jacalz:
@hajimehoshi Are you perhaps the right one to ask for regenerating `gl`?

hajimehoshi:
Well, however, I'm now pretty busy now. Anyone can regenerate `gl` and send a PR. Thanks,

Jacalz:
> Well, however, I'm now pretty busy now. Anyone can regenerate `gl` and send a PR. Thanks,

I was not sure how to do it. Saw the documentation afterwards and did it myself instead :)

PR #346, opened 2022-01-14
v3.3/glfw: expose native wayland methods

rajveermalviya:
exposes native wayland & EGL methods.

https://github.com/glfw/glfw/blob/df8d7bc892937a8b0f7c604c92a9f64f383cf48c/include/GLFW/glfw3native.h#L420

### DIFF
--- a/v3.3/glfw/native_linbsd_wayland.go
+++ b/v3.3/glfw/native_linbsd_wayland.go
@@ -1,0 +1,47 @@
+// +build linux,wayland freebsd,wayland netbsd,wayland openbsd,wayland
+
+package glfw
+
+//#include <stdlib.h>
+//#define GLFW_EXPOSE_NATIVE_WAYLAND
+//#define GLFW_EXPOSE_NATIVE_EGL
+//#define GLFW_INCLUDE_NONE
+//#include "glfw/include/GLFW/glfw3.h"
+//#include "glfw/include/GLFW/glfw3native.h"
+import "C"
+
+func GetWaylandDisplay() *C.struct_wl_display {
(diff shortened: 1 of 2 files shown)


rajveermalviya:
@dmitshur PTAL

v3.3/glfw/native_linbsd_wayland.go
```diff
@@ -0,0 +1,47 @@
+// +build linux,wayland freebsd,wayland netbsd,wayland
```
Jacalz: Shouldn't this contain `openbsd,wayland` as well?

rajveermalviya: yes, updated

Jacalz: Thanks.

dmitshur: Now that the new file has `openbsd,wayland` added, it seems `native_linbsd_x11.go` should also be updated so its build constraint is `openbsd,!wayland`.

Jacalz: Good catch. That must have been missed when the OpenBSD support was added in.

rajveermalviya: > Now that the new file has openbsd,wayland added, it seems native_linbsd_x11.go should also be updated so its build constraint is openbsd,!wayland.

done.

Jacalz (approved):
Sorry for the late review. Looks good to me.

dmitshur:
Thanks.

PR #350, opened 2022-04-06
v3.3/glfw: update to GLFW 3.3.7

Jacalz:
See https://www.glfw.org/changelog.html for list of changes.
There are a bunch of bug fixes for X11 and Wayland but also a few on Windows and macOS.

This is working fine on my Fedora 35 laptop but I am getting a compilation warning:
```
In file included from ../glfw/v3.3/glfw/c_glfw_lin.go:20:
../glfw/v3.3/glfw/glfw/src/x11_window.c:30: warning: "_GNU_SOURCE" redefined
   30 | #define _GNU_SOURCE
      | 
```


Jacalz:
I have opened an issue about the compilation warning over at https://github.com/glfw/glfw/issues/2076. This does not impact the compilation or running of binaries but it is perhaps not very elegant to have the warning there.

elmindreda:
Should we be defining `_GNU_SOURCE` conditionally? Not sure why it's being defined on the command-line.

Jacalz:
I'm not sure how that works in C, but I expect it to at least not be defined twice. I don't have a better answer, sorry.

dmitshur:
@Jacalz It looks like `build.go` is where the `_GNU_SOURCE` is being defined the first time. Now that GLFW 3.3.7 is taking the responsibility for defining that, it seems we can drop the definition from `build.go` and avoid the redefinition warning. Also see https://github.com/glfw/glfw/issues/2076#issuecomment-1120469603.

dmitshur (approved):
LGTM. Thanks for preparing this, and sorry it took a while to get to it.



See comment above about the `_GNU_SOURCE` redefinition warning. I think it's okay to either submit this as is and follow up on the warning in another PR, or if you'd like you can try updating this PR. What would you prefer @Jacalz?

Jacalz:
> See comment above about the `_GNU_SOURCE` redefinition warning. I think it's okay to either submit this as is and follow up on the warning in another PR, or if you'd like you can try updating this PR. What would you prefer @Jacalz?

I think I should be able to get it fixed here in this PR. Thanks for looking at this and tracking down the source of that issue.

Jacalz:
I have updated this to now remove the `_GNU_SOURCE` (I hope it was the correct fix) but I now get this warning when building:
```
In file included from ../glfw/v3.3/glfw/c_glfw_lin.go:20:
../glfw/v3.3/glfw/glfw/src/x11_window.c: In function ‘waitForData’:
../glfw/v3.3/glfw/glfw/src/x11_window.c:77:32: warning: implicit declaration of function ‘ppoll’; did you mean ‘poll’? [-Wimplicit-function-declaration]
   77 |             const int result = ppoll(fds, count, &ts, NULL);
```

hajimehoshi:
Interesting. This call actually exists in the 3.3-stable branch but not in the master branch.

`_GNU_SOURCE` is still needed for `ppoll` (https://linux.die.net/man/2/ppoll), right?

Jacalz:
Strange. What should be the way forward then?

dmitshur:
To me, the warning that `_GNU_SOURCE` is defined twice seems more harmless. If you're not able to find a solution for `ppoll` problem easily, I'd suggest let's revert 692d67bc69fe52c59d22b0fcad6adff63e3c2c74 and just merge this 3.3.7 update as is, then file an issue about finding the best path to eliminate the warning (using what we've learned from here).

hajimehoshi:
I agree that the duplicated `_GNU_SOURCE` is much more harmless.

Jacalz:
I agree too. Sounds like a good plan. The commit has been reverted and I will make sure to open a new issue about the warning. Feel free to merge :)

Jacalz:
I opened https://github.com/go-gl/glfw/issues/351 to track it. However, I am starting to think that it actually might be an upstream issue now that we don't define `_GNU_SOURCE` ourselves.

hajimehoshi (approved):

dmitshur:
Thanks!

PR #354, opened 2022-05-15
v3.3/glfw: minor code cleanups

Jacalz:
This removes an if-statement that was unnecessary and removes an unnecessary integer conversion.
This only touches the `v3.3/glfw` package because I don't have any good code to test the others with.


hajimehoshi (approved):
LGTM

Jacalz:
Interestingly enough, the compiler actually doesn't seem to be able to optimize away the if-statement. The new code should be ever so slightly smaller and faster (far from noticeable though). https://go.godbolt.org/z/E34h4K36n

dmitshur:
Thanks.

PR #356, opened 2022-06-20
fix _GNU_SOURCE redefined warning

### LINKED ISSUE #351: v3.3/glfw: _GNU_SOURCE is defined twice, emits a warning
As of upstream glfw v3.3.7 (and https://github.com/go-gl/glfw/pull/350), it is now setting `_GNU_SOURCE` along with our `v3.3/glfw/build.go` setting it for us. This means that the C-compiler files a warning about it being defined twice.
```
In file included from ../glfw/v3.3/glfw/c_glfw_lin.go:20:
../glfw/v3.3/glfw/glfw/src/x11_window.c:30: warning: "_GNU_SOURCE" redefined
   30 | #define _GNU_SOURCE
      | 
```

The easy fix is to remove `-D_GNU_SOURCE` from the `build.go` file but that unfortunatelly makes a another warning pop up.
```
In file included from ../glfw/v3.3/glfw/c_glfw_lin.go:20:
../glfw/v3.3/glfw/glfw/src/x11_window.c: In function ‘waitForData’:
../glfw/v3.3/glfw/glfw/src/x11_window.c:77:32: warning: implicit declaration of function ‘ppoll’; did you mean ‘poll’? [-Wimplicit-function-declaration]
   77 |             const int result = ppoll(fds, count, &ts, NULL);
```

We need to find a better way of solving this.

cornelk:
Closes #351

In upstream the _GNU_SOURCE and poll changes introduced this:

https://github.com/glfw/glfw/commit/9f73e9afa34b1c0902023d5d80be1c0f954135e5
https://github.com/glfw/glfw/commit/adc202d2c3182ca6ad8344624941e56d8e0bc493

### DIFF
--- a/v3.3/glfw/build.go
+++ b/v3.3/glfw/build.go
@@ -26,8 +26,8 @@ package glfw
 // Linux Build Tags
 // ----------------
 // GLFW Options:
-#cgo linux,!wayland CFLAGS: -D_GLFW_X11 -D_GNU_SOURCE
-#cgo linux,wayland CFLAGS: -D_GLFW_WAYLAND -D_GNU_SOURCE
+#cgo linux,!wayland CFLAGS: -D_GLFW_X11
+#cgo linux,wayland CFLAGS: -D_GLFW_WAYLAND
 
 // Linker Options:
 #cgo linux,!gles1,!gles2,!gles3,!vulkan LDFLAGS: -lGL
--- a/v3.3/glfw/c_glfw_lin.go
+++ b/v3.3/glfw/c_glfw_lin.go
@@ -15,9 +15,9 @@ package glfw
 	#include "glfw/src/wayland-xdg-shell-client-protocol.c"
 #endif
 #ifdef _GLFW_X11
+	#include "glfw/src/x11_window.c"
 	#include "glfw/src/x11_init.c"
 	#include "glfw/src/x11_monitor.c"
-	#include "glfw/src/x11_window.c"
 	#include "glfw/src/glx_context.c"
 #endif
 #include "glfw/src/linux_joystick.c"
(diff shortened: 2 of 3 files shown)


Jacalz:
You might want to contribute this change upstream. I'm not sure modifying the Glfw source in this way is the best option. It will be removed as soon as Glfw is updated again.

cornelk:
@Jacalz this has already been fixed in upstream (i guess for 3.4): https://github.com/glfw/glfw/commit/bb9d699ae66b2bdc8718995ba13c57c9c8e59602 - I am not sure what to do here

Jacalz:
I think it would be best to get a PR up to fix the issue for their next 3.3.x release. Otherwise it's just going to be undone work when we update to the next release in that series.

That would mean targeting the 3.3-stable branch: https://github.com/glfw/glfw/tree/3.3-stable.

cornelk:
@Jacalz I have a feeling that this will not get accepted ( https://github.com/glfw/glfw/issues/2133#issuecomment-1162311920 ) but I will submit it there

dmitshur:
@cornelk I think what I wrote in https://github.com/go-gl/glfw/issues/351#issuecomment-1126811178 applies. If the problem can be reproduced with the GLFW C source code (independently of the Go bindings), then it's reasonable to target the fix there. The report in glfw/glfw#2133 refers to the Go bindings, which is suboptimal for that issue tracker. We should investigate that on our side in #351.

cornelk:
@Jacalz i believe i have found the proper fix now by changing the import order, i see no warning anymore (i am on Linux) using:

```
cd v3.3/glfw/
go clean -cache
go build ./...
```

@dmitshur thanks for the tip, i checked and could build glfw 3.3 stable without any warning.

Jacalz:
Cool. Nice work 👍

Jacalz (approved):
Nice work. Thanks for the fix.

dmitshur:
Thanks.

PR #358, opened 2022-07-12
v3.3/glfw: regenerate Wayland protocols

Jacalz:
This should hopefully fix/unblock https://github.com/fyne-io/fyne/pull/3131.

PR #361, opened 2022-07-23
v3.3: Update glfw to v3.3.8

Jacalz:
This PR updates the version of glfw to 3.3.8 for the v3.3 module. The new release includes a bunch of new bug fixes: https://www.glfw.org/changelog.


dmitshur:
Thanks. One question so far, the rest looks good.

It looks like the wayland files were previously in v3.3/glfw/glfw/src directory, which this PR updates, but it also adds similar files in the v3.3 directory. Is that working as expected, or should we only have one set of those files in one directory?

dmitshur:
Friendly ping @Jacalz, see question above when you get a chance. Thanks.

Jacalz:
Sorry. I forgot to answer and now I’ve been on a short vacation. I think it might just have been a mistake on my end but will have a look in a few days time.

dmitshur:
No problem, and thanks!

Jacalz:
I have removed the files now. I don't know why they were there in the first place bu I suppose that they might have been some leftover from https://github.com/go-gl/glfw/pull/358 in my local Git tree.

dmitshur (approved):
Thanks.

PR #363, opened 2022-08-23
Add Bazel build support

garymm:
For v3.3, Linux-X11 and macOS.


Jacalz:
No offence, but all Go projects that I know of just build using the regular Go tooling. This seems to me like something that adds extra complexity each time we update GLFW. Do we really want to carry the burden of maintaining this?

garymm:
No offense taken. My project uses Bazel, and I agree we're unusual.
Given that I did the work for my own project, I thought I'd share this in the hopes that it'd be useful to others. 
I think it will only need to be updated when a source file is added or deleted from v3.3.
Up to you whether that is too much of a maintenance burden.

garymm:
@Jacalz could you please either close this, merge it, or LMK what you'd like me to do to get this ready to be merged?

Jacalz:
I don't think it is up to me in the end. Better check with @dmitshur first

PR #366, opened 2022-10-17
v3.3/glfw: Regenerate Wayland protocols on Ubuntu 20.04

Jacalz:
Wayland protocols are backwards compatible, but not necessarily forward compatible. Ubuntu 20.04 has version 1.18 so let's regenerate with that to give us a wider degree of OS support. This means that we now support version >= 1.18 instead of >= 1.20 (from my Fedora 36 install).


dmitshur (approved):
Thanks.

PR #367, opened 2022-10-18
Address conflict between clang in osxcross and what latest MacOS SDK

Bluebugs:
Latest MacOS SDK is using an attribute that confuse the clang provided by osxcross. As this attribute is just going to generate a warning at build time and seems to be of low value, getting rid of it seems like the easiest path forward that I can think of.

### DIFF
--- a/v3.1/glfw/glfw/src/internal.h
+++ b/v3.1/glfw/glfw/src/internal.h
@@ -56,6 +56,8 @@
  #error "No supported client library selected"
 #endif
 
+#define NS_FORMAT_ARGUMENT(A)
+
 // Disable the inclusion of the platform glext.h by gl.h to allow proper
 // inclusion of our own, newer glext.h below
 #define GL_GLEXT_LEGACY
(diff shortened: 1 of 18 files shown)


Bluebugs:
Not to sure if this would be the logical way to go at this problem for the go-glfw project. Let me know if you can think of an alternate solution.

Jacalz:
Is this perhaps something that should be added to upstream glfw instead? Seems like something that could affect them as well.

My personal opinion is that you patch upstream glfw instead of modifying the `v3.3/glfw/glfw/src` files and then ignore patching up the `v3.0`, `v3.1`, and `v3.2` folders all together

v3.1/glfw/glfw/src/internal.h
```diff
@@ -56,6 +56,8 @@
  #error "No supported client library selected"
 #endif
 
+#define NS_FORMAT_ARGUMENT(A)
```
Jacalz, line 59: Per my previous comment, this is a file synced with upstream glfw releases. We likely wasn't to add this upstream instead

Jacalz (changes requested):

Bluebugs:
Upgrading to llvm v14 solve the problem. It is easier than maintaining this patch. Abandoning it then.

Issue #379, opened 2023-10-10
`(*glfw.Window).SetIcon` doesn't consider strides


hajimehoshi:
I tested this with `github.com/go-gl/glfw/v3.3/glfw v0.0.0-20210410170116-ea3d685f79fb`.

```go
	pix := make([]byte, 4*32*16)
	for j := 0; j < 16; j++ {
		for i := 0; i < 16; i++ {
			pix[4*(i+32*j)] = 0xff
			pix[4*(i+32*j)+1] = 0
			pix[4*(i+32*j)+2] = 0
			pix[4*(i+32*j)+3] = 0xff
		}
		for i := 16; i < 32; i++ {
			pix[4*(i+32*j)] = 0
			pix[4*(i+32*j)+1] = 0xff
			pix[4*(i+32*j)+2] = 0
			pix[4*(i+32*j)+3] = 0xff
		}
	}
	icon := &image.NRGBA{
		Pix:    pix,
		Stride: 4*32,
		Rect:   image.Rect(0, 0, 16, 16),
	}
	window.SetIcon([]image.Image{icon})
```

The icon should be a red square, but the actual icon was a square with red and green stripes.

I believe `glfw.CreateCursor` has the same issue.

PR #380, opened 2023-10-30
Fix image handling

egonelbre:
Each commit is a separate change and reviewing the whole diff is more difficult.

- v3.1,v3.2,v3.3: fix image conversion in CreateCursor and CreateIcon
- v3.0,v3.1,v3.2: make util.go consistent
- all: code formatting

### DIFF
--- a/v3.1/glfw/util.go
+++ b/v3.1/glfw/util.go
@@ -8,36 +8,21 @@ import "C"
 import (
 	"image"
 	"image/draw"
-	"reflect"
-	"unsafe"
 	"runtime"
 )
 
 func glfwbool(b C.int) bool {
-	if b == C.int(True) {
-		return true
-	}
-	return false
+	return b == C.int(True)
 }
 
 func bytes(origin []byte) (pointer *uint8, free func()) {
 	n := len(origin)
-
 	if n == 0 {
 		return nil, func() {}
 	}
 
-	data := C.malloc(C.size_t(n))
-
-	dataSlice := *(*[]byte)(unsafe.Pointer(&reflect.SliceHeader{
-		Data: uintptr(data),
-		Len:  n,
-		Cap:  n,
-	}))
-
-	copy(dataSlice, origin)
-
-	return &dataSlice[0], func() { C.free(data) }
+	ptr := C.CBytes(origin)
+	return (*uint8)(ptr), func() { C.free(ptr) }
 }
 
 // imageToGLFW converts img to be compatible with C.GLFWimage.
--- a/v3.1/glfw/util.go
+++ b/v3.1/glfw/util.go
@@ -6,8 +6,11 @@ package glfw
 import "C"
 
 import (
+	"image"
+	"image/draw"
 	"reflect"
 	"unsafe"
+	"runtime"
 )
 
 func glfwbool(b C.int) bool {
@@ -35,4 +38,25 @@ func bytes(origin []byte) (pointer *uint8, free func()) {
 	copy(dataSlice, origin)
 
 	return &dataSlice[0], func() { C.free(data) }
+}
+
+// imageToGLFW converts img to be compatible with C.GLFWimage.
+// It may reference the underlying pixels buffer in img.
+func imageToGLFW(img image.Image) (r C.GLFWimage, free func()) {
+	b := img.Bounds()
+
+	r.width = C.int(b.Dx())
+	r.height = C.int(b.Dy())
+
+	if m, ok := img.(*image.NRGBA); ok && m.Stride == b.Dx()*4 {
+		r.pixels = (*C.uchar)(&m.Pix[0])
+		return r, func() { runtime.KeepAlive(m) }
+	}
+
+	m := image.NewNRGBA(image.Rect(0, 0, b.Dx(), b.Dy()))
+	draw.Draw(m, m.Bounds(), img, b.Min, draw.Src)
+
+	pix, free := bytes(m.Pix)
+	r.pixels = (*C.uchar)(pix)
+	return r, free
 }
--- a/v3.2/glfw/util.go
+++ b/v3.2/glfw/util.go
@@ -5,6 +5,12 @@ package glfw
 //#include "glfw/include/GLFW/glfw3.h"
 import "C"
 
+import (
+	"image"
+	"image/draw"
+	"runtime"
+)
+
 func glfwbool(b C.int) bool {
 	if b == C.int(True) {
 		return true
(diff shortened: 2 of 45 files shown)


egonelbre:
CC @hajimehoshi can you take a look whether this fixes your issue?

hajimehoshi:
I think we should separate the PR into formatting and fixing the issue

hajimehoshi:
> your issue

You mean https://github.com/go-gl/glfw/issues/379?

egonelbre:
> I think we should separate the PR into formatting and fixing the issue

I used separate commits for each of the changes, because they do end up conflicting somewhat. Unfortunately, github doesn't do stacked diffs that well, so people need to click each commit separately for reviewing. And managing separate PR for each change is annoyingly time consuming.

Note, the formatting was done automatically via `gofmt -w -s`.

> You mean https://github.com/go-gl/glfw/issues/379?

Yes.

hajimehoshi:
I don't know this project's policy how we treat PRs, so I'd like to defer this to other reviewers.

> Yes.

Thanks!

v3.1/glfw/util.go
```diff
@@ -6,33 +6,40 @@ package glfw
...
+
+	r.width = C.int(b.Dx())
+	r.height = C.int(b.Dy())
 
-	dataSlice := *(*[]byte)(unsafe.Pointer(&reflect.SliceHeader{
-		Data: uintptr(data),
-		Len:  n,
-		Cap:  n,
-	}))
+	if m, ok := img.(*image.NRGBA); ok && m.Stride == b.Dx() && b.Min.X == 0 {
```
hajimehoshi:
```go
m.Stride == b.Dx() * 4
```
?

egonelbre: sigh... indeed

egonelbre: also thank you

(the sigh was more due to my mistake)

egonelbre: fixed

v3.2/glfw/util.go
```diff
@@ -21,3 +23,23 @@ func bytes(origin []byte) (pointer *uint8, free func()) {
...
 	return (*uint8)(ptr), func() { C.free(ptr) }
 }
+
+func imageToGLFW(img image.Image) (r C.GLFWimage, free func()) {
+	b := img.Bounds()
+
+	r.width = C.int(b.Dx())
+	r.height = C.int(b.Dy())
+
+	if m, ok := img.(*image.NRGBA); ok && m.Stride == b.Dx()*4 && b.Min.X == 0 {
```
hajimehoshi: Do we really need `b.Min.X == 0`?

v3.2/glfw/util.go
```diff
@@ -21,3 +23,23 @@ func bytes(origin []byte) (pointer *uint8, free func()) {
...
+		pix, free := bytes(m.Pix[b.Min.Y*m.Stride : b.Max.Y*m.Stride])
```
hajimehoshi: Does this really work when `b.Min.Y` and/or `b.Max.Y` is negative? I thought just `m.Pix` should work.

https://pkg.go.dev/image#NRGBA

> // Pix holds the image's pixels, in R, G, B, A order. The pixel at
> // (x, y) starts at Pix[(y-Rect.Min.Y)*Stride + (x-Rect.Min.X)*4].
> Pix [][uint8](https://pkg.go.dev/builtin#uint8)

v3.2/glfw/util.go
egonelbre: I think it does, but I'm not completely confident.

egonelbre: Yeah, it can be done without that check as well.

v3.2/glfw/util.go
egonelbre: Yeah, you seem right. For some reason, I was thinking that it didn't subslice the pix when making a subimage.

I guess the better approach would be:

```
if m, ok := img.(*image.NRGBA); ok && m.Stride == b.Dx()*4 {
	start := m.PixOffset(m.Rect.Min.X, m.Rect.Min.Y)
	end := m.PixOffset(m.Rect.Max.X, m.Rect.Max.Y)
	pix, free := bytes(m.Pix[start:end])
```

Then it's clear that the ranges are calculated correctly.

egonelbre: So, yeah, start ends up always `0`, however, end might be something different, because Subimage doesn't limit the length https://cs.opensource.google/go/go/+/refs/tags/go1.21.3:src/image/image.go;l=430

v3.1/glfw/util.go
```diff
@@ -6,33 +6,41 @@ package glfw
...
-	dataSlice := *(*[]byte)(unsafe.Pointer(&reflect.SliceHeader{
-		Data: uintptr(data),
-		Len:  n,
-		Cap:  n,
-	}))
+	if m, ok := img.(*image.NRGBA); ok && m.Stride == b.Dx()*4 {
+		end := m.PixOffset(m.Rect.Max.X, m.Rect.Max.Y)
+		pix, free := bytes(m.Pix[:end])
+		r.pixels = (*C.uchar)(pix)
+		return r, free
```
hajimehoshi: Do we need to copy pixels? i.e., what about:

```go
return (*C.uchar)(unsafe.Pointer(&m.Pix[0])), func() { runtime.KeepAlive(m) }
```

?

dmitshur:
> I don't know this project's policy how we treat PRs, so I'd like to defer this to other reviewers.

In lieu of GitHub having better support for stacked changes, I think it's fine to review the individual commits, and once everything's ready, use the "Rebase and merge" strategy to merge it. That way the 3 individual commits will appear in the history of the main branch.

v3.1/glfw/util.go
egonelbre: I did consider it, but I wasn't sure how the glfw image will be later used. If it's not going to be used in other scenarios later down the line then it should be fine. And as far as I understand, the functions using images seemed such that they aren't called every frame, so the allocation overhead probably won't matter.

Either way, I'll add the optimization.

egonelbre: Done

hajimehoshi: Thanks. Is it intended to have 4 commits in this PR?

egonelbre: Yes.

hajimehoshi:
Let's merge ebfd54ae46714e2172ff4f5f43ee8ec308383c0d and e8d0e1b7832f2baeff55146e6b7345fa3a6c3527 into one commit ~~since ebfd54ae46714e2172ff4f5f43ee8ec308383c0d adds copying bytes and e8d0e1b7832f2baeff55146e6b7345fa3a6c3527 removes it.~~ for simplicity, but what do you think?

hajimehoshi (approved):
LGTM, thanks!

PR #384, opened 2023-11-01
Fix cgo argument has Go pointer to unpinned Go pointer

### LINKED ISSUE #383: Panic when creating cursor after latest changes
In e1dbfaf, new logic surrounding the handling of cursor images was introduced. That logic leads to panics when `glfw.CreateCursor` is called. Those panics happen on at least Windows and MacOS.

The panic message is:

```
panic: runtime error: cgo argument has Go pointer to unpinned Go pointer
```

The relevant part of the stack trace is:

```
panic({0x100beb300?, 0xc00051a0f0?})
        /usr/local/Cellar/go/1.21.0/libexec/src/runtime/panic.go:914 +0x21f
github.com/go-gl/glfw/v3.3/glfw.CreateCursor.func1(0x1021431f0?, 0xb, 0xa)
        /Users/kaioreilly/go/pkg/mod/github.com/go-gl/glfw/v3.3/glfw@v0.0.0-20231031225837-d1c54e5847d0/input.go:500 +0x3d
github.com/go-gl/glfw/v3.3/glfw.CreateCursor({0x1021431f0, 0xc00044d640}, 0x20?, 0x1031005b8?)
        /Users/kaioreilly/go/pkg/mod/github.com/go-gl/glfw/v3.3/glfw@v0.0.0-20231031225837-d1c54e5847d0/input.go:500 +0x86
goki.dev/goosi/driver/desktop.(*cursorImpl).Set(0x102cdcd00, {0x102145138, 0x1029ae6c8})
        /Users/kaioreilly/goki/goosi/driver/desktop/cursor.go:48 +0x246
goki.dev/gi/v2/gi.(*EventMgr).SetCursor(0xc00019c1d0?, 0x2151a10?)
        /Users/kaioreilly/goki/gi/gi/eventmgr.go:596 +0xc9
```

`desktop/cursor.go:48` is just calling `gc := glfw.CreateCursor(ci.Image, h.X, h.Y)`, where `ci.Image` is an `image.Image` that it has just opened from an   embed FS PNG file.

egonelbre:
It seems that avoiding copy in image wasn't valid and I didn't test the change.

This contains two commits, one to fix the issue and the other adds an example that makes testing easier.

Fixes #383

### DIFF
--- a/v3.1/glfw/util.go
+++ b/v3.1/glfw/util.go
@@ -8,7 +8,6 @@ import "C"
 import (
 	"image"
 	"image/draw"
-	"runtime"
 )
 
 func glfwbool(b C.int) bool {
@@ -26,22 +25,22 @@ func bytes(origin []byte) (pointer *uint8, free func()) {
 }
 
 // imageToGLFW converts img to be compatible with C.GLFWimage.
-// It may reference the underlying pixels buffer in img.
 func imageToGLFW(img image.Image) (r C.GLFWimage, free func()) {
 	b := img.Bounds()
 
 	r.width = C.int(b.Dx())
 	r.height = C.int(b.Dy())
 
+	var pixels []byte
 	if m, ok := img.(*image.NRGBA); ok && m.Stride == b.Dx()*4 {
-		r.pixels = (*C.uchar)(&m.Pix[0])
-		return r, func() { runtime.KeepAlive(m) }
+		pixels = m.Pix[:m.PixOffset(m.Rect.Min.X, m.Rect.Max.Y)]
+	} else {
+		m := image.NewNRGBA(image.Rect(0, 0, b.Dx(), b.Dy()))
+		draw.Draw(m, m.Bounds(), img, b.Min, draw.Src)
+		pixels = m.Pix
 	}
 
-	m := image.NewNRGBA(image.Rect(0, 0, b.Dx(), b.Dy()))
-	draw.Draw(m, m.Bounds(), img, b.Min, draw.Src)
-
-	pix, free := bytes(m.Pix)
+	pix, free := bytes(pixels)
 	r.pixels = (*C.uchar)(pix)
 	return r, free
 }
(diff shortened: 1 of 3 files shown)


egonelbre:
CC @hajimehoshi

hajimehoshi:
Thanks,

>  an example that makes testing easier.

We already have go-gl/example, so is it OK to add examples in this repository? (/CC @dmitshur )

hajimehoshi:
Is [`runtime.Pinner`](https://pkg.go.dev/runtime#Pinner) available instead of copying by the way?

NVM: This was introduced in Go 1.21, which is too early to adopt.

hajimehoshi:
Yet another suggestion is https://pkg.go.dev/runtime/cgo#Handle

egonelbre:
go-gl/example only tests for one of the glfw versions, and it seems slightly nicer to be able to use an example directly from this repo.

> Yet another suggestion is https://pkg.go.dev/runtime/cgo#Handle

cgo.Handle serves a different purpose -- it's to avoid passing a pointer to C, but in this case glfw c code needs to access the byte buffer.

hajimehoshi:
> go-gl/example only tests for one of the glfw versions, and it seems slightly nicer to be able to use an example directly from this repo.

As I don't know the policy, I'd like to defer the decision to other reviewers.

> cgo.Handle serves a different purpose -- it's to avoid passing a pointer to C, but in this case glfw c code needs to access the byte buffer.

Ah OK, I was misunderstanding. Thanks!

hajimehoshi:
By the way, if you are in hurry, please separate a PR into fixing the issue and adding the example.

egonelbre:
Sure. I removed the example from here.

PS: not in a hurry, but I did introduce a significant bug.

egonelbre:
And moved the example commit here https://github.com/go-gl/glfw/pull/385

hajimehoshi (approved):
LGTM, thanks!

PR #385, opened 2023-11-02
Add easily runnable examples

egonelbre:
Add minimal examples that can be used to test basic functionality.

### DIFF
--- a/v3.1/glfw/testdata/customcursor/main.go
+++ b/v3.1/glfw/testdata/customcursor/main.go
@@ -1,0 +1,52 @@
+package main
+
+import (
+	"image"
+	"image/color"
+	"runtime"
+
+	"github.com/go-gl/glfw/v3.1/glfw"
+)
+
+func init() {
+	// This is needed to arrange that main() runs on main thread.
+	// See documentation for functions that are only allowed to be called from the main thread.
+	runtime.LockOSThread()
+}
+
+func main() {
+	err := glfw.Init()
+	if err != nil {
+		panic(err)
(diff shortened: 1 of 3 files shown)


dmitshur (commented):
I think the distinction between example and test is that a test is intended for developers of a package, and an example is intended for users. Since this is primarily a test, I suggest using that term.

Specifically, how about renaming v3.3/glfw/example/main.go to v3.3/glfw/test/customcursor.go, and adding a "//go:build test" constraint, so it doesn't show up when listing packages while still being easy to run by a developer with `go run customcursor.go`.

Another comment, I think it's generally fine for us to maintain and develop only v3.3 (or whichever is the latest and actively used), and leave v3.2 and v3.1 as is. This is mostly for the future, it's not a big deal either way in this PR. Thanks.

egonelbre:
Ah, maybe in that case I'll put it in `glfw/testdata/customcursor/main.go`? AFAIR `testdata` is hidden by default and usually I avoid build tags, because they mean bunch of tooling won't work on the files.

dmitshur:
That seems fine too.

> usually I avoid build tags, because they mean bunch of tooling won't work on the files

My hope is that work on issues like golang/go#29202 and golang/go#57979 will make this less of a problem over time.

egonelbre:
I did the changes.

But, yes, improvements at the tooling level would be great.

egonelbre:
@dmitshur tiny ping

dmitshur (approved):
Thanks.

PR #388, opened 2023-12-18
Update upstream GLFW to v3.3.9

### LINKED ISSUE #378: Any way to try the glfw master branch to use libdecor ?
Hello,

As we already discussed earlier for Fyne.io project, Wayland is badly supported at this time as the decoration should be made client side.

GLFW seems to now manage libdecor: 
https://github.com/glfw/glfw/blob/3eaf1255b29fdf5c2895856c7be7d7185ef2b241/docs/intro.dox#L166C6-L166C27

But, of course, there is no release from their repository to try this with Go :)

Have you got a manual to help me to test it with go-gl/glfw?

I know that you use some scripts to get/pull versions, but that not successful last time I tried.

A basic, TLDR would be appreciate ;)

Thanks a lot !

Jacalz:
Upstream GLFW release: https://github.com/glfw/glfw/releases/tag/3.3.9
Includes many bug fixes and improved Wayland support.

Fixes #378


Jacalz:
I am getting this error when compiling with `-tags wayland`. Does anyone have an idea for what package is missing?
Tried with installing `wayland-protocols-devel` and `libdecor-devel` (on top of existing `wayland-devel`) on Fedora 39 and nothing makes a difference.
```
In file included from ./glfw/src/internal.h:190,
                 from ./glfw/src/context.c:30,
                 from ../../../glfw/v3.3/glfw/c_glfw.go:4:
./glfw/src/wl_platform.h:59:10: fatal error: xdg-shell-client-protocol.h: No such file or directory
   59 | #include "xdg-shell-client-protocol.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

dmitshur:
Thanks for getting this started.

If you haven't already, see if there are hints in previous work on this that you can find [here](https://github.com/go-gl/glfw/issues?q=%22xdg-shell-client-protocol%22).
Maybe more to consider at https://www.glfw.org/docs/latest/compile.html#compile_deps_wayland.
Also CC @santiagocezar, @andydotxyz in case you have ideas.

Jacalz:
The problem turned out to be that the script for regenerating wayland protocols was broken due to the GitHub repository no longer existing. I didn't see the error when grabbing the new GLFW upstream but managed to see it now. Everything is working after fixing the script.

Fyne Wayland before this update:
image

Fyne Wayland after this update:
image

andydotxyz:
Oh this is very exciting 😀

scripts/generate-wayland-protocols.sh
```diff
@@ -9,7 +9,7 @@ if [ "$GLGLFW_PATH" = "" ]; then
     exit 1
```
dmitshur, line 11: Consider adding [`set -e`](https://explainshell.com/explain?cmd=set+-e) at the top. That would've helped catch the issue with `git clone` failing silently.

v3.3/glfw/glfw/include/dummy.go
```diff
@@ -1,4 +1,3 @@
-//go:build required
```
dmitshur: We probably shouldn't drop this line.

I think the fix is to update `grab-upstream.sh`—it currently writes just `// +build required`. The new `go:build` lines were probably added by gofmt in the past, but it's better to update the source `grab-upstream.sh` to include them to begin with.

All supported versions of Go long since understand `go:build`, so `+build` should be fine to drop. Up to you.

dmitshur (approved):
Thanks. See two inline comments, otherwise LGTM.

v3.3/glfw/glfw/include/dummy.go
Jacalz: I thought I had fixed them all but it seems not. Will update the script to add `go:build`.

scripts/generate-wayland-protocols.sh
Jacalz: Nice idea. Fixed that in the latest commit

Jacalz:
@dmitshur Do you want re-approve or is fine to merge?

scripts/grab-upstream.sh
```diff
@@ -24,6 +24,8 @@ excludes=(
   --exclude=glfw-*/tests
 )
 
+set -e
```
dmitshur, line 27: Consider moving this up to be at the top of script (on line 3, similar to how it's done in generate-wayland-protocols.sh). The motivation is to follow the pattern of setting it at the very beginning of the script, before doing anything else, and make it easier for code readers to see it's there.

scripts/grab-upstream.sh
```diff
@@ -91,6 +95,7 @@ EOF
   grab-upstream v3.3/glfw
 
   cat > v3.3/glfw/glfw/deps/dummy.go <<EOF
+//go:build required
```
dmitshur: (nit) There's a trailing space at the end of this line (it's not there in the other two above). Since `gofmt` would delete it, let's not include it.

dmitshur:
@Jacalz I didn't realize you were waiting on me. (I appreciate how Gerrit has the [attention set](https://gerrit-review.googlesource.com/Documentation/user-attention-set.html) concept which makes it easier to see whose turn it is; it'd be nice if GitHub had something to help with that.)

It still LGTM, though I've added minor nits while here. Please feel free to merge after addressing them. Thanks.

dmitshur (approved):

scripts/grab-upstream.sh
dmitshur: While here, you could also add it to `glfw_tree_rebuild.sh`, so that all 3 scripts have it. No reason for that one not to.

Jacalz:
Over at Fyne we require an approval before merging and any new commits after an approval dismiss the approval. Just unsure how you do things here. Will fix up and then merge :)

PR #390, opened 2023-12-29
v3.3/glfw: remove 'immediately' from internal error message

dmitshur:
An internal error is an error that should never happen, but it can still happen when there is a bug in the code. When it happens, glfw prints a suggestion of reporting the bug, as that way we can learn about the bug and fix it.

The word "immediately" may create a sense of urgency, but in fact these bug reports are not very time sensitive, at least no more than any other issue. Relax the language a bit.

Also, as commonly done, print these errors to stderr rather than stdout.

### DIFF
--- a/v3.3/glfw/error.go
+++ b/v3.3/glfw/error.go
@@ -8,6 +8,7 @@ import "C"
 import (
 	"fmt"
 	"log"
+	"os"
 )
 
 // ErrorCode corresponds to an error code.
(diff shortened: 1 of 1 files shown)


hajimehoshi (approved):
LGTM

Just out of curiosity, why not using stderr instead of stdout?

dmitshur:
> Just out of curiosity, why not using stderr instead of stdout?

That's just what it was before. I agree switching to stderr is an improvement. Done in commit [`da841ce2`](https://github.com/go-gl/glfw/compare/f0b0de189eb6d3f14dc9c782ebaa5e1038add131..da841ce248d3fffac2690c7d563103498337fa14).

Thanks.

PR #394, opened 2024-02-29
update README with required packages for Wayland

BinarSkugga:
Haven't used every components but these packages are required for a basic window to show up in Wayland.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
 * GLFW C library source is included and built automatically as part of the Go package. But you need to make sure you have dependencies of GLFW:
 	* On macOS, you need Xcode or Command Line Tools for Xcode (`xcode-select --install`) for required headers and libraries.
 	* On Ubuntu/Debian-like Linux distributions, you need `libgl1-mesa-dev` and `xorg-dev` packages.
-	* On CentOS/Fedora-like Linux distributions, you need `libX11-devel libXcursor-devel libXrandr-devel libXinerama-devel mesa-libGL-devel libXi-devel libXxf86vm-devel` packages.
+	* On CentOS/Fedora-like Linux distributions, you need `libX11-devel libXcursor-devel libXrandr-devel libXinerama-devel mesa-libGL-devel libXi-devel libXxf86vm-devel` packages. If you develop with wayland, you need `libxkbcommon-devel wayland-devel mesa-libGL-devel`.
 	* On FreeBSD, you need the package `pkgconf`. To build for X, you also need the package `xorg`; and to build for Wayland, you need the package `wayland`.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -5,7 +5,7 @@
 * GLFW C library source is included and built automatically as part of the Go package. But you need to make sure you have dependencies of GLFW:
 	* On macOS, you need Xcode or Command Line Tools for Xcode (`xcode-select --install`) for required headers and libraries.
 	* On Ubuntu/Debian-like Linux distributions, you need `libgl1-mesa-dev` and `xorg-dev` packages.
-	* On CentOS/Fedora-like Linux distributions, you need `libX11-devel libXcursor-devel libXrandr-devel libXinerama-devel mesa-libGL-devel libXi-devel libXxf86vm-devel` packages.
+	* On CentOS/Fedora-like Linux distributions, you need `libX11-devel libXcursor-devel libXrandr-devel libXinerama-devel mesa-libGL-devel libXi-devel libXxf86vm-devel` packages. If you develop with wayland, you need `libxkbcommon-devel wayland-devel mesa-libGL-devel`.
```
dmitshur: It seems a bit unclear if this means "you _only_ need" or "you _also_ need". Since "mesa-libGL-devel" is repeated, I'm guessing it's not the latter?

It might be more clear to reuse the phrasing from below:

"On CentOS/Fedora-like Linux distributions, you need [...common packages...]. To build for X, also need [...extra packages for X...]; and to build for Wayland, you also need [...extra packages for Wayland...]."

dmitshur (commented):
Thanks.

As a side note, I'm starting to wonder if we should instead (or in addition) link to the upstream documentation on this at https://www.glfw.org/docs/latest/compile_guide.html#compile_deps, which I suspect applies well to the Go bindings for GLFW too.

README.md
Jacalz: It is also worth noting that `libdecor-devel` can be useful as an optional dependency for better titlebars.

BinarSkugga: Sorry for the delay, work became hectic. Should be clearer !

README.md
```diff
@@ -5,7 +5,7 @@
 * GLFW C library source is included and built automatically as part of the Go package. But you need to make sure you have dependencies of GLFW:
 	* On macOS, you need Xcode or Command Line Tools for Xcode (`xcode-select --install`) for required headers and libraries.
 	* On Ubuntu/Debian-like Linux distributions, you need `libgl1-mesa-dev` and `xorg-dev` packages.
-	* On CentOS/Fedora-like Linux distributions, you need `libX11-devel libXcursor-devel libXrandr-devel libXinerama-devel mesa-libGL-devel libXi-devel libXxf86vm-devel` packages.
+	* On CentOS/Fedora-like Linux distributions, you need `mesa-libGL-devel`. To build for X11, you'll need `libX11-devel libXcursor-devel libXrandr-devel libXinerama-devel libXi-devel libXxf86vm-devel` packages; and for Wayland you'll need `libxkbcommon-devel wayland-devel`; if you're using Gnome with Wayland, you should also install `libdecor-devel`.
```
Jacalz, line 8: I don't think you should explicitly call out Gnome here. Libdecor solves decorations for more systems and it's not really about the developer's experience but the user running the released application. Maybe just write that it is an optional and recommended dependency for improved window docorarions on Wayland?

Jacalz (changes requested):

PR #398, opened 2024-03-07
scripts: Drop support for grabbing v3.2 upstream

### LINKED ISSUE #397: scripts: Grab upstream fails to apply patch for v3.2
When updating `v3.3/glfw`, I have commented out the grabbing of upstream for `v3.2/glfw` because it fails to apply the patch. I looked into it a bit but didn't really know it behaves like that. It would be nice to have this fixed to make updates smoother. Or maybe v3.2 is dead in the water and we should just drop it from the script?

```
patching file README.md
Hunk #1 FAILED at 173.
1 out of 1 hunk FAILED
patching file src/x11_platform.h
Hunk #1 succeeded at 122 (offset -19 lines).
patching file src/x11_window.c
Hunk #1 succeeded at 993 (offset 1 line).
```

Jacalz:
As noted in the issue, v3.2 no longer manages to be set up with the script but it is EOL anyway so let's just remove it.

Fixes #397


dmitshur (approved):
Thanks.

PR #212, opened 2017-11-26
Fix build for wayland

r0l1:
This is a work in progress pull request to get the wayland issue fixed as soon as possible.
This is based on #199 work.

Right now the build progress fails with:

```
# github.com/go-gl/glfw/v3.2/glfw
In file included from ./c_glfw_linbsd.go:16:0:
./glfw/src/wayland-relative-pointer-unstable-v1-client-protocol.c:34:35: error: redefinition of ‘types’
 static const struct wl_interface *types[] = {
                                   ^~~~~
In file included from ./c_glfw_linbsd.go:15:0:
./glfw/src/wayland-pointer-constraints-unstable-v1-client-protocol.c:37:35: note: previous definition of ‘types’ was here
 static const struct wl_interface *types[] = {
                                   ^~~~~
```

Any hints how to fix this?

Thanks!

PR #399, opened 2024-04-23
Remove warning/errors about undefined functions

### LINKED ISSUE #359: v3.3/glfw: Warnings when building using -tags wayland
## New warnings:

This is now the warning message we get after https://github.com/go-gl/glfw/pull/399 has been merged to fix Fedora 40 now treating implicit declarations as an error due to their stricter security:
```
In file included from ../../c_glfw_lin.go:10:
../../glfw/src/wl_window.c:29:9: warning: "_GNU_SOURCE" redefined
   29 | #define _GNU_SOURCE
      |         ^~~~~~~~~~~
```

This should be fixed so we get no warnings. Removing `_GNU_SOURCE` being defined will make the build fail on Fedora 40.

## Old warnings:

I am getting these errors when building for Wayland. Opening the issue here in case there are issues on our end like last time.
```
In file included from ../../vendor/github.com/go-gl/glfw/v3.3/glfw/c_glfw_lin.go:9:
../../vendor/github.com/go-gl/glfw/v3.3/glfw/glfw/src/wl_window.c: In function ‘createTmpfileCloexec’:
../../vendor/github.com/go-gl/glfw/v3.3/glfw/glfw/src/wl_window.c:115:10: warning: implicit declaration of function ‘mkostemp’; did you mean ‘mkstemp’? [-Wimplicit-function-declaration]
  115 |     fd = mkostemp(tmpname, O_CLOEXEC);
      |          ^~~~~~~~
      |          mkstemp
../../vendor/github.com/go-gl/glfw/v3.3/glfw/glfw/src/wl_window.c: In function ‘waitForData’:
../../vendor/github.com/go-gl/glfw/v3.3/glfw/glfw/src/wl_window.c:266:32: warning: implicit declaration of function ‘ppoll’; did you mean ‘poll’? [-Wimplicit-function-declaration]
  266 |             const int result = ppoll(fds, count, &ts, NULL);
../../vendor/github.com/go-gl/glfw/v3.3/glfw/glfw/src/wl_window.c: In function ‘_glfwPlatformGetClipboardString’:
../../vendor/github.com/go-gl/glfw/v3.3/glfw/glfw/src/wl_window.c:1866:11: warning: implicit declaration of function ‘pipe2’; did you mean ‘pipe’? [-Wimplicit-function-declaration]
 1866 |     ret = pipe2(fds, O_CLOEXEC);
      |           ^~~~~
      |           pipe
```

metal3d:
Fixes #359

features.h needs `_GNU_SOURCE` to be set to allow stdlib.h to declare the functions.

### DIFF
--- a/v3.3/glfw/build.go
+++ b/v3.3/glfw/build.go
@@ -26,7 +26,7 @@ package glfw
 // Linux Build Tags
 // ----------------
 // GLFW Options:
-#cgo linux,!wayland CFLAGS: -D_GLFW_X11 -D_GNU_SOURCE
+#cgo linux,!wayland CFLAGS: -D_GLFW_X11
 #cgo linux,wayland CFLAGS: -D_GLFW_WAYLAND -D_GNU_SOURCE
 
 // Linker Options:
--- a/v3.3/glfw/build.go
+++ b/v3.3/glfw/build.go
@@ -26,8 +26,8 @@ package glfw
 // Linux Build Tags
 // ----------------
 // GLFW Options:
-#cgo linux,!wayland CFLAGS: -D_GLFW_X11
-#cgo linux,wayland CFLAGS: -D_GLFW_WAYLAND
+#cgo linux,!wayland CFLAGS: -D_GLFW_X11 -D_GNU_SOURCE
+#cgo linux,wayland CFLAGS: -D_GLFW_WAYLAND -D_GNU_SOURCE
 
 // Linker Options:
 #cgo linux,!gles1,!gles2,!gles3,!vulkan LDFLAGS: -lGL


metal3d:
fixes #359

dmitshur (dismissed):
Seems okay to me.

We had this in v3.2 but didn't propagate it forward to v3.3:

https://github.com/go-gl/glfw/blob/a69d953ea14231b2f19998f0b434b6c1e6c6f710/v3.2/glfw/build.go#L25-L26

(It was added there in #212 and 691ee1b84c51ae625ed1aafebc915d3db3d63d66.)

Thanks. Giving a chance to @Jacalz to also take a look before merging.

Jacalz (changes requested):
Thanks. I am however getting this warning on both X11 and Wayland compilations on Fedora Silverblue 39. Is there any way that we can avoid getting warnings about multiple definitions?
```
In file included from ../../c_glfw_lin.go:10:
../../glfw/src/wl_window.c:29:9: warning: "_GNU_SOURCE" redefined
   29 | #define _GNU_SOURCE
      |         ^~~~~~~~~~~
```
Could it be that the file that previously was giving the warnings should have a `#define _GNU_SOURCE` instead and these build changes should be removed?

EDIT: I tried my suggested fix and it does not seem to improve anything :(

dmitshur (commented):
Looking more, I found we removed these from v3.3 in #356 (to fix #351). So I agree it seems this needs more investigation. Also see past discussion in https://github.com/glfw/glfw/issues/2133 and https://github.com/glfw/glfw/issues/2076.

Jacalz:
Hmm. I have one idea. Given that those warnings on Wayland now are errors after the linker changes in Fedora 40 (I've confirmed what @metal3d saw), how about we just enable GNU_SOURCE for Wayland builds?

It's obviously just a short-term fix but it does improve things quite a bit. Anyone using Fedora 40 can then compile the binary (three errors become one warning) and everyone else gets one warning instead of three. Seems like a win until we manage to get all the warnings to disappear?

metal3d:
I agree @Jacalz, this is critical on Fedora 40 and probably with others distributions that are up to date. Without this fix, it's impossible to build any Fyne application with Wayland support.

A warning isn't problematic as it's only an already defined directive. IMHO this shouldn't be a problem, even if the warning is fixed, later, in the glfw project.

Jacalz:
Yeah, but given that there is no compilation problem on X11, I see not point in adding a new warning just for the sake of it. Would you mind keeping it enabled only for Wayland?

metal3d:
FYI, I made the change to only set the directive to compile on Wayland. It is a very important fix that allows Fyne application to compile on Fedora 40.

Even if it's temporary, this fix is highly necessary.

Jacalz (approved):
Sorry for the late review. I was away running a half-marathon during the weekend. This looks great. Thanks for the fix.

Issue #375, opened 2023-03-30
v3.4 and supporting GLFW_MOUSE_PASSTHROUGH


J7mbo:
Hi folks! I'm displaying a transparent window overlay to highlight things on the screen (a red rectangle) and I'd really like to be able to "still click through" to things underneath this rectangle. I found that glfw itself has: `GLFW_MOUSE_PASSTHROUGH` from 2020 I think. I couldn't find this in the code in here, so I assume it's not implemented yet.

Would it be possible to add this without too much effort? If it's just a redirect from to a C function or something, I'd be happy to learn and follow the pattern you do for other calls, but I may need support on knowing exactly what to do.

Please let me know what you think and if you'd be willing to support this?

J7mbo:
It _looks_ like this is [only for v3.4](https://www.glfw.org/docs/3.4/window_guide.html#GLFW_MOUSE_PASSTHROUGH_hint). So I guess my request is actually about supporting version v3.4. I'll change the title of the issue.

J7mbo:
Additionally, it's probable that supporting an entirely new version would be a tonne of work and it's not fair for me to request this just for a single window hint.

However would a maintainer / supporter be willing to suggest or advise here were I to attempt to do this myself, just for a single window hint, by following the patterns set by this library for the other versions? 

Maybe the PR could even begin the work for supporting v3.4 and if it's not complicated I could contribute with the rest if it's just "passthrough" function calls. :)

J7mbo:
Closing as it looks like this isn't interesting for anyone. I'll figure out another way.

J7mbo:
Looks this [this issue](https://github.com/go-gl/glfw/issues/393) will solve this.

DownloadableFox:
Sorry if I shouldn't be typing here considering it's closed, I tried to reach you out on Twitter but it seems you don't have DMs enabled. I have been experiencing the same issue with a personal project that I have pause whilst waiting for the release of GLFW 3.4. I was looking to use the `GLFW_MOUSE_PASSTHROUGH` but of course, it is not yet available. 

I was wondering if you have found any alternative to this so far?

J7mbo:
Hey @DownloadableFox, nope not found anything else yet. Just waiting for this MR. There's no instructions there that I can see yet as I'm on a Mac.

I'll re-open as it's being actively worked on.

J7mbo:
Okay I got this working.

[...]

PR #400, opened 2024-05-05
GLFW v3.4

### LINKED ISSUE #272: v3.4/glfw: PostEmptyEvent crashes if called first on non-main thread on macOS
This is the `golang.org/go-gl/glfw/v3.3/glfw` tracking bug for upstream GLFW issue glfw/glfw#1649. It can be reproduced with the following Go program:

```Go
package main

import (
	"log"
	"runtime"

	"github.com/go-gl/glfw/v3.3/glfw"
)

func init() { runtime.LockOSThread() }

func main() {
	err := glfw.Init()
	if err != nil {
		log.Fatalln(err)
	}
	defer glfw.Terminate()

	go glfw.PostEmptyEvent() // call PostEmptyEvent in non-main thread

	select {}
}
```

Running it on macOS 10.15.3 with Xcode 11.3.1 produces the following crash:

<details><br>

```
$ go run .                                     
2020-02-22 13:50:10.100 m[22196:92162] *** Assertion failure in +[NSUndoManager _endTopLevelGroupings], /BuildRoot/Library/Caches/com.apple.xbs/Sources/Foundation/Foundation-1674.114/Foundation/Misc.subproj/NSUndoManager.m:363
2020-02-22 13:50:10.100 m[22196:92162] *** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: '+[NSUndoManager(NSInternal) _endTopLevelGroupings] is only safe to invoke on the main thread.'
*** First throw call stack:
(
	0   CoreFoundation                      0x00007fff394778ab __exceptionPreprocess + 250
	1   libobjc.A.dylib                     0x00007fff6f731805 objc_exception_throw + 48
	2   CoreFoundation                      0x00007fff394a0d10 +[NSException raise:format:arguments:] + 88
	3   Foundation                          0x00007fff3bb99241 -[NSAssertionHandler handleFailureInMethod:object:file:lineNumber:description:] + 191
	4   Foundation                          0x00007fff3bad7d5e +[NSUndoManager(NSPrivate) _endTopLevelGroupings] + 440
	5   AppKit                              0x00007fff365b016c -[NSApplication run] + 864
	6   m                                   0x00000000040b0046 _glfwPlatformPostEmptyEvent + 54
	7   m                                   0x000000000405cc90 runtime.asmcgocall + 112
)
libc++abi.dylib: terminating with uncaught exception of type NSException
SIGABRT: abort
PC=0x7fff70be67fa m=3 sigcode=0
```

</details>

To make progress here, we need to resolve glfw/glfw#1649 first, wait for a new GLFW version with the fix included to be released, and then pull in the fixed GLFW version into `golang.org/go-gl/glfw/v3.3/glfw`.

### LINKED ISSUE #393: v3.4/glfw: add support for GLFW version 3.4 (and future 3.4.x versions)
The initial release of GLFW v3.4 was released a few days ago, see https://github.com/glfw/glfw/releases/tag/3.4. I'm opening this issue to keep track of the work of adding the necessary support here.

Geo25rey:
## Overview

This is a rough implementation of go-glfw using GLFW v3.4. I have done some basic testing on Linux Wayland and XWayland. I highly recommend reviewing by commit considering the nature and size of this PR.

I am also looking for people to test the builds and functionality on the various platforms listed below. I can personally continue testing on Linux and if I have extra time, MacOS Arm.

Closes: #393
Potentially fixes: #272 @dmitshur 
Requesting Review from: @Jacalz 

## Rundown by commit
- [v3.3 copy - starting point](https://github.com/go-gl/glfw/commit/77aeccf1b006c31a69801ccb778e834853d6fea6)
  - 1-to-1 copy of the v3.3 Go code excluding the GLFW C library itself
- [update version number in imports](https://github.com/go-gl/glfw/commit/8e37f8f7a617f7bfdbaa45b67e397bd827274c0b)
  - changes all instances of "v3.3" in the imports to "v3.4"
- [glfw source gen for v3.4+](https://github.com/go-gl/glfw/commit/2d26992c34ba2d326898677fad62407ae240033e)
  - adds a new script that uses the CMake build system in the upstream GLFW repo to generate dependency headers
    - for now this is just wayland, but this should be applicable for others if added in the future
  - includes the source code gathered from [GLFW 3.4](https://github.com/glfw/glfw/tree/3.4)
- [update cgo linux C source file list](https://github.com/go-gl/glfw/commit/a7c03f1a25b0ea2eca2787b5efa54c412af84ff9)
  - updates the path to the new and renamed source files from upstream GLFW
  - build specifically tested on Linux
- [disable unsupported setting of window icon on wayland](https://github.com/go-gl/glfw/commit/3ff1abd63dd68554fe150150fde619fb86110396)
  - GLFW will crash the process with SIGABRT when calling `C.glfwSetWindowIcon()` under a Wayland environment because it's not supported
  - there might be a better way to handle GLFW_FEATURE_UNAVAILABLE errors. open to suggestions
  - error implementation: [v3.4/glfw/glfw/src/wl_window.c:2227](https://github.com/go-gl/glfw/commit/2d26992c34ba2d326898677fad62407ae240033e#diff-74f4782c75f1a54e227cb1834c8c3771899784a1c390edf5773fb5366104056dR2227)

### Preamble

I used the short Custom Cursor program for testing. It would probably be a good idea to test with other programs, as well. Does anyone have such test programs?

It is especially important to get more test programs for Wayland since more features than just setting the window icon are not supported by GLFW and/or the Wayland protocols yet, and I'm not sure of other implications of the SIGABRT behavior previously mentioned. For example, I could dynamically add some element to my window at runtime, which could crash my program with a cryptic abort message with no other context. At best, developers using go-glfw can report it here, and at worst a subset of users will report the issue to downstream projects that use go-glfw and the downstream developers may not know what to do with the error message.

### `cd v3.4/glfw; go run `[`testdata/customcursor/main.go`](https://github.com/go-gl/glfw/commit/77aeccf1b006c31a69801ccb778e834853d6fea6#diff-5fa04a85ad541b0e369e9dbc75fb1bdf9b4c75f99b8ac62e40a43d632f8d9174)
- [ ] Linux
  - [ ] Wayland
    - [x] OpenGL (`-tags linux,wayland`)
    - [ ] GLESv1 (`-tags linux,wayland,gles1`)
      - doesn't compile on Arch Linux with latest Mesa drivers
    - [x] GLESv2 (`-tags linux,wayland,gles2`)
    - [ ] GLESv3 (`-tags linux,wayland,gles3`)
      - doesn't compile on Arch Linux with latest Mesa drivers
    - [x] Vulkan (`-tags linux,wayland,vulkan`)
  - [ ] X11
    - [x] OpenGL (`-tags linux`)
    - [ ] GLESv1 (`-tags linux,gles1`)
      - doesn't compile on Arch Linux with latest Mesa drivers
    - [x] GLESv2 (`-tags linux,gles2`)
    - [ ] GLESv3 (`-tags linux,gles3`)
      - doesn't compile on Arch Linux with latest Mesa drivers
    - [x] Vulkan (`-tags linux,vulkan`)
- [ ] BSD
  - [ ] FreeBSD
    - [ ] Wayland (`-tags freebsd,wayland`)
    - [ ] X11 (`-tags freebsd`)
  - [ ] NetBSD
    - [ ] Wayland (`-tags netbsd,wayland`)
    - [ ] X11 (`-tags netbsd`)
  - [ ] OpenBSD
    - [ ] Wayland (Is this unsupported? There is no tag set for this.)
    - [ ] X11 (`-tags openbsd`)
- [ ] Mac
  - [ ] Arm (I can get to this at some point, but if someone would like to get to first, feel free. I'm not sure when I'll have time for it)
    - [x] opengl32 (`-tags darwin`)
    - [ ] GLESv2 (`-tags darwin,gles2`)
  - [ ] x86
    - [ ] opengl32 (`-tags darwin`)
    - [ ] GLESv2 (`-tags darwin,gles2`)
- [ ] Windows (not sure if both 10 & 11 need to be tested but better safe than sorry, right?)
  - [ ] 10
    - [ ] opengl32 (`-tags windows`)
    - [ ] GLESv2 (`-tags windows,gles2`)
  - [ ] 11
    - [ ] opengl32 (`-tags windows`)
    - [ ] GLESv2 (`-tags windows,gles2`)

## TODO before merge
- [ ] Fix BSD build
- [ ] Fix Windows build
- [ ] Update window hints
- [ ] Cleanup commit history (if needed)
- [ ] Make issue for Feature Unavailable and Unimplemented errors not cause Go panic. https://github.com/go-gl/glfw/pull/400#issuecomment-2118868433
- [ ] Make issue for what default build tags should be on Linux (X11, Wayland, or both)


dmitshur:
Thanks for working on this.

> I used the short Custom Cursor program for testing. It would probably be a good idea to test with other programs, as well. Does anyone have such test programs?

It's not much, but there are two small example programs in https://github.com/go-gl/example.

---

Out of the box, it fails to link on GOOS=darwin GOARCH=arm64 with:

<details>

```
# command-line-arguments
/usr/local/go/pkg/tool/darwin_arm64/link: running clang failed: exit status 1
Undefined symbols for architecture arm64:
  "__glfwPlatformFreeModule", referenced from:
      __glfwTerminateNull in 000002.o
      __glfwTerminateOSMesa in 000002.o
      _terminate in 000002.o
      __glfwInitVulkan in 000002.o
      __glfwTerminateVulkan in 000002.o
      __glfwInitOSMesa in 000002.o
      __glfwTerminateCocoa in 000003.o
      ...
  "__glfwPlatformGetModuleSymbol", referenced from:
      __glfwInitVulkan in 000002.o
      _glfwGetInstanceProcAddress in 000002.o
      __glfwInitOSMesa in 000002.o
      __glfwInitOSMesa in 000002.o
      __glfwInitOSMesa in 000002.o
      __glfwInitOSMesa in 000002.o
      __glfwInitOSMesa in 000002.o
      __glfwInitOSMesa in 000002.o
      __glfwInitOSMesa in 000002.o
      ...
  "__glfwPlatformLoadModule", referenced from:
      __glfwInitVulkan in 000002.o
      __glfwInitOSMesa in 000002.o
      __glfwLoadLocalVulkanLoaderCocoa in 000003.o
      __glfwInitEGL in 000003.o
      __glfwCreateContextEGL in 000003.o
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

</details>

That doesn't happen if this line is added to `c_glfw_darwin.go`:

```diff
 #include "glfw/src/cocoa_time.c"
+#include "glfw/src/posix_module.c"
 #include "glfw/src/posix_thread.c"
```

And with that at least basic functionality seems to work okay on macOS.

Geo25rey:
> It's not much, but there are two small example programs in https://github.com/go-gl/example.

I'll go ahead and try these then.

> `c_glfw_darwin.go`:
> ```diff
>  #include "glfw/src/cocoa_time.c"
> +#include "glfw/src/posix_module.c"
>  #include "glfw/src/posix_thread.c"
> ```

Cool I'll add this in.

By the way, does this fix #272 for you? I think the upstream fix is included in GLFW v3.4.

Geo25rey:
Forgot to mention, @dmitshur can you test with `-tags darwin,gles2` as well?

Jacalz:
Without the changes in https://github.com/go-gl/glfw/pull/399 (and also defining `_GNU_SOURCE` for X11 compared to v3.3), I can't compile this on Fedora 40 due to implicit declarations now being treated as errors. Test results after that change:

I can replicate the same gles1 and gles3 failures on Fedora 40 (Wayland and X11) but I get the same error on v3.3 so that's fine? Compiling with `-tags gles` does work though.

Isn't Glfw v3.4 supposed to be able to switch between Wayland and X11 at runtime? If that is the case, should we enable  that by default? Maybe have x11 and wayland tags for when you only want a single backend?

Jacalz:
FreeBSD X11 seems to get this error:
image

Geo25rey:
> FreeBSD X11 seems to get this error: image

Does `go build -tags freebsd` work?

Geo25rey:
> Without the changes in #399 (and also defining `_GNU_SOURCE` for X11 compared to v3.3), I can't compile this on Fedora 40 due to implicit declarations now being treated as errors. Test results after that change:

I can define `_GNU_SOURCE`, but I get this warning, which could be an error in Fedora?
```
In file included from ./c_glfw_lin.go:9:
./glfw/src/wl_window.c:27: warning: "_GNU_SOURCE" redefined
   27 | #define _GNU_SOURCE
```

> I can replicate the same gles1 and gles3 failures on Fedora 40 (Wayland and X11) but I get the same error on v3.3 so that's fine? Compiling with `-tags gles` does work though.

I think it's fine too. To my understanding, GLES v1 & v3 are just different implementations of the OpenGL spec. So, it's probably good keeping them in, but they might be hard to test on bleeding edge distros like Arch and Fedora.

Also, I believe `-tags gles` is ignored when building for Linux & BSD. It would just choose the default which is `-lGL` implementation, right?

> Isn't Glfw v3.4 supposed to be able to switch between Wayland and X11 at runtime? If that is the case, should we enable that by default? Maybe have x11 and wayland tags for when you only want a single backend?

I think it's best to keep X11 as the default to make the transition easier from v3.3 to v3.4, but if GLFW can now choose Wayland or X11 at runtime, maybe we can add support for this when both `x11` and `wayland` tags are included.

Geo25rey:
I just pushed support for applying both X11 & Wayland on Linux with (-tags x11,wayland). I will add to BSD later, but I want to get the build working standalone on BSD first.

Jacalz:
> Does go build -tags freebsd work?

That won't make any difference as I was building on FreeBSD. The GOOS value for the native operating system is always added automatically and the same goes for GOARCH. If you build your app on Linux for example, you don't need to pass `-tags linux`.

Jacalz:
> I can define _GNU_SOURCE, but I get this warning, which could be an error in Fedora?

No, it is not an error in Fedora. It is a drawback of the solution. See https://github.com/go-gl/glfw/issues/359, we need a better solution but right now that is better than not being able to compile. Implicit declarations are emitted (which is treated as an error on Fedora 40 due to stricter security settings) if your don't force _GNU_SOURCE to on but if you do, you get a warning that it is defined twice.

> I think it's best to keep X11 as the default to make the transition easier from v3.3 to v3.4, but if GLFW can now choose Wayland or X11 at runtime, maybe we can add support for this when both x11 and wayland tags are included.

I'm not personally of doing that approach. Wayland is the future and I think we should enable the runtime switching support if we can. However, it is up to @dmitshur at the end of the day.

metal3d:
> I can define `_GNU_SOURCE`, but I get this warning, which could be an error in Fedora?
> 
> ```
> In file included from ./c_glfw_lin.go:9:
> ./glfw/src/wl_window.c:27: warning: "_GNU_SOURCE" redefined
>    27 | #define _GNU_SOURCE
> ```

Yes, this is a warning due to the manual definition inside GLFW source code. I don't understand why it's defined without a `#ifndef` - it could be a mistake or a wanted behavior.

v3.4/glfw/build.go
```diff
@@ -0,0 +1,57 @@
...
+
+#cgo !gles2,darwin LDFLAGS: -framework OpenGL
+#cgo gles2,darwin LDFLAGS: -lGLESv2
+
+// Linux Build Tags
+// ----------------
+// GLFW Options:
+#cgo linux,!wayland CFLAGS: -D_GLFW_X11
+#cgo linux,x11 CFLAGS: -D_GLFW_X11
+#cgo linux,wayland CFLAGS: -D_GLFW_WAYLAND
```
metal3d, line 33: See my comment on top, is it OK that `_GNU_SOURCE` is now globally set?

v3.4/glfw/build.go
```diff
@@ -0,0 +1,57 @@
+package glfw
+
+/*
+#cgo CFLAGS: -Iglfw/include -D_GNU_SOURCE
```
metal3d, line 4: So then, `_GNU_SOURCE` is now globally set?

v3.4/glfw/build.go
Jacalz: We probably don't want it set but I couldn't compile at all without it.

Geo25rey:
@metal3d I think `_GNU_SOURCE` is better to be set for all platforms since GLFW already handles cross-platform functionality. Ideally, `_GNU_SOURCE` should be set directly before the required `#include`, but it seems that there is an issue with the order the C source files are included into Go code. I think this is the most robust solution rather than hoping someone doesn't reorder the C source files in the future.

Geo25rey:
> Wayland is the future

@Jacalz I completely agree, but this isn't about ideologies. This is about what is best for the users of this library.

> we should enable the runtime switching support if we can.

I have added support for this on Linux like I said previously (`-tags x11,wayland`). Right now, I just have it setup as an opt-in feature rather than replacing the X11 default.

Jacalz:
> @Jacalz I completely agree, but this isn't about ideologies. This is about what is best for the users of this library.

My point has nothing to do with ideologies. You are taking things out of context and I don't agree that being X11-only would be helpful in the long run. Enabling Wayland-only by default would be stupid but supporting dynamic switching between the two is a good compromise that we should try to make use of now that we can.

Enabling runtime switching by default is the best path forward given how quickly Wayland is getting adopted and some distros like Fedora are even disabling Xorg support (XWayland works yes, but you see the direction we are heading). Making the switch to enable Wayland support by default now is going to be the better option as a future-proof setting and that switch will be much harder to make down the line. Remember that v3.4 is an entirely new module import and it would be problematic to suddenly enable Wayland support and require an extra dependency when compiling later in the lifecycle of that existing in the repository. If we are to stay up to date with the times, that support needs to be enabled from the start. If you still want to have X11-only, I see no reason why people can't still use v3.3 given that it still will be maintained some time forward. 

We have seen in Fyne that the community cares very much for Wayland support to be enabled and Flathub even marks an X11-only application as unsafe nowadays (https://flathub.org/apps/io.github.jacalz.rymdport). Having dynamic runtime switching gets us both X11 and Wayland support and a lot more people will actually use it instead of having to enable an obscure build tag combination.

J7mbo:
Hey @Geo25rey, given that Mac --> opengl32 is checked here, I tried building on an M3 with:

> go build -tags darwin

And then running `./customcursor`.

I had to comment out:

> window.SetIcon([]image.Image{whiteTriangle})

in  v3.4/glfw/testdata/customcursor/main.go:36, as it was complaining about Cocoa:

> go-gl/glfw: internal error: an invalid error was not accepted by the caller: ErrorCode(65548): Cocoa: Regular windows do not have icons on macOS
go-gl/glfw: Please report this in the Go package issue tracker.
panic: ErrorCode(65548): Cocoa: Regular windows do not have icons on macOS

Given the above being commented out, the mouse passthrough I was looking for worked (with floating), after adding the Hint to `window.go`, so thanks a lot for that! :)

Geo25rey:
@J7mbo Thx for testing this out according to https://github.com/LWJGL/lwjgl3/issues/695, this commit https://github.com/glfw/glfw/commit/9a87c2a4b49808d3666c7b093b77e8512fa23908 introduced 2 new errors affecting macOS and Wayland, so I think we should pass both `GLFW_FEATURE_UNAVAILABLE` and `GLFW_FEATURE_UNIMPLEMENTED` back as go errors instead of panicking

Geo25rey:
@J7mbo I just looked at your issue https://github.com/go-gl/glfw/issues/375#issuecomment-2118842844 and it looks like all the hints are out of date or have out of date descriptions so I will add them from here https://www.glfw.org/docs/3.4/window_guide.html#window_hints_wnd

v3.4/glfw/not_wayland.go
```diff
@@ -0,0 +1,6 @@
+//go:build !((linux && wayland) || (freebsd && wayland) || (netbsd && wayland) || (openbsd && wayland))
+// -build linux,wayland freebsd,wayland netbsd,wayland openbsd,wayland
```
dmitshur, line 2: (I thought I left this review comment before, but it doesn't seem to show up. Trying it again.)

This line seems to be malformed. Maybe it was intended to be `// +build`. I think it'd be fine not to include it at all, since all versions of Go since Go 1.17 can handle build constraints specified via just the `//go:build` lines.

dmitshur (commented):
Thanks for getting it this far. I've been using this locally on macOS and haven't spotted any more issues on this platform and in my usage.

---

> Forgot to mention, @dmitshur can you test with `-tags darwin,gles2` as well?

I gave it a try:

```
$ go run -tags=gles2 .
# github.com/go-gl/example/trygles2
/usr/local/go/pkg/tool/darwin_arm64/link: running clang failed: exit status 1
ld: warning: ignoring duplicate libraries: '-lGLESv2'
ld: library 'GLESv2' not found
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

It's the same error as I get with glfw/v3.3. I've never tried to use OpenGL ES on macOS, and I don't think it's supported. Maybe it is on iOS, but I don't think go-gl/glfw (or upstream GLFW) tries to support iOS at this time.

https://www.glfw.org/docs/latest/window_guide.html#GLFW_CONTEXT_VERSION_MAJOR_hint includes:

> macOS: The OS only supports core profile contexts for OpenGL versions 3.2 and later.

---

> I have added support for this on Linux like I said previously (`-tags x11,wayland`). Right now, I just have it setup as an opt-in feature rather than replacing the X11 default.

Thanks for working on this and getting this work at all, that's great. There was a question of what would be preferred default behavior.

My general approach to answering these kinds of questions is to try to follow the upstream GLFW decisions whenever viable, with a balance of satisfying constraints on the Go side. Since GLFW v3.4 [changed its default](https://www.glfw.org/docs/latest/news.html#use_wayland_removed) compared to GLFW v3.3 from X11 (with Wayland selected at compile-time) to X11+Wayland selected at runtime, if we can make that the default for glfw/v3.4, that seems optimal to me. That is, based on my current understanding, having what is currently selected with `-tags=x11,wayland` be selected with `-tags=` (no custom tags) instead seems optimal for the eventual glfw/v3.4 version. I'm certainly open to being convinced otherwise if there are reasons this doesn't work well.

If we think more discussion is needed, I also think it's fine to leave it as is and make it a follow up issue instead of trying to resolve it as part of this PR.

DownloadableFox:
Are there any extra steps or requirements for building in Windows? I have tried to build in both Windows 10 and 11 with different results, but failing in both. I tried the same steps in my Arch Linux machine and it worked first try.

Geo25rey:
Thx for testing on Windows @DownloadableFox 

Please share the build errors you have found so they can be fixed.

v3.4/glfw/not_wayland.go
Geo25rey: Cool. I'll go ahead and fix this. I'll look at the docs. Do you have a link to the relevant docs?

v3.4/glfw/build.go
Geo25rey: If no response to https://github.com/go-gl/glfw/pull/400#issuecomment-2098661241, I'll just mark this as resolved.

v3.4/glfw/build.go
Geo25rey: If no response to https://github.com/go-gl/glfw/pull/400#issuecomment-2098661241, I'll just mark this as resolved.

Geo25rey:
@dmitshur I have added a TODO list for items I'd like to get done before merging this in the PR description https://github.com/go-gl/glfw/pull/400#issue-2279777601

> If we think more discussion is needed, I also think it's fine to leave it as is and make it a follow up issue instead of trying to resolve it as part of this PR.

I agree. We should probably pull in someone from the Fyne team, as well. I'm sure they'd have some useful insights on this.

DownloadableFox:
> Thx for testing on Windows @DownloadableFox
> 
> Please share the build errors you have found so they can be fixed.

I don't have my Windows 11 computer with me at the moment, but this is the output from Windows 10:
image

It appears to be several lines of linker problems. If you want the full log I'd be happy to share it with you, though it's pretty big. I tried building v3.3 and it built with no problems under the same environment. Currently using Go 1.22.1 and MINGW64 through MSYS2.

Jacalz:
> I agree. We should probably pull in someone from the Fyne team, as well. I'm sure they'd have some useful insights on this.

I am part of the Fyne team. That's where my insights and opinions on Wayland come from

Geo25rey:
@DownloadableFox please provide the full plain text output either here in a < summary > < details > HTML tag or in a public gist.

Geo25rey:
@Jacalz How about we discuss in a voice call in the Fyne Discord server? Do you have availability this weekend? 

Also, I'd like to include someone from the Fyne team that is not opinionated about Wayland itself to hopefully prevent any stalemates, if that is alright with you.

Jacalz:
> @Jacalz How about we discuss in a voice call in the Fyne Discord server? Do you have availability this weekend?

Voice call sounds like a good plan. I won't be available this weekend unfortunately but the next one or next week should hopefully be a better option.

> Also, I'd like to include someone from the Fyne team that is not opinionated about Wayland itself to hopefully prevent any stalemates, if that is alright with you.

Sure. If you say so.

Geo25rey:
@Jacalz how about 1pm EDT on Monday, May 3?

Jacalz:
That works well for me. Should we check on Discord if someone more from our team is available?

Geo25rey:
Yeah sounds good. @Jacalz 

Can you ask around?

DownloadableFox:
> @DownloadableFox please provide the full plain text output either here in a < summary > < details > HTML tag or in a public gist.

As a correction to my previous comment, both Windows 11 and 10 share the same output. I might have not looked at it correctly last time I tried to build. Here's the log:

https://gist.github.com/DownloadableFox/33c16e3e26566a1eb1252454a0b1e87d

EDIT: Ended up making it a Gist instead due to formatting issues.

DownloadableFox:
> > @DownloadableFox please provide the full plain text output either here in a < summary > < details > HTML tag or in a public gist.
> 
> As a correction to my previous comment, both Windows 11 and 10 share the same output. I might have not looked at it correctly last time I tried to build. Here's the log:
> 
> https://gist.github.com/DownloadableFox/33c16e3e26566a1eb1252454a0b1e87d
> 
> EDIT: Ended up making it a Gist instead due to formatting issues.

I was actually able to fix this after a couple of hours tampering with the code. I was able to build and run in both Windows 10 and 11 by adding this line to `g_glfw_windows.go`.

```
#include "glfw/src/win32_module.c"
```

Jacalz:
> Yeah sounds good. @Jacalz 
> 
> 
> 
> Can you ask around?

Absolutely. Andrew (`@andydotxyz` here on GitHub), the founder and team leader of Fyne, is on board as well 👍

J7mbo:
Did this happen in the end? Any update?

Jacalz:
The meeting on the third of June hasn't happened yet

DownloadableFox:
Any updates on this?

Jacalz:
We had a discussion and came to the conclusion that X11 + Wayland is the way to go. It is the default upstream and what the whole ecosystem is pushing for. It is a change that will be harder to implement in the future.

dmitshur:
Checking in on the status. @Geo25rey, would you like to keep pushing this forward? If so, there are some unresolved review comments that need to be addressed, otherwise this seems close. I've been using this locally for a while without uncovering new issues.

v3.4/glfw/not_wayland.go
dmitshur: Build constraints are documented at https://pkg.go.dev/cmd/go#hdr-Build_constraints.

Geo25rey:
@dmitshur I can work on it tomorrow. I have a patch for Windows support to finish and I need to update the Linux default build tags.

Jacalz:
Are you able to finish this sometime soon? It would be great to get this merged when we can :)

J7mbo:
Folks is there any update on this? Is @Geo25rey alone on this one or could someone knowledgable enough support?

sedyh:
+1 for this.
I can't use glfw 3.3 since fractional scaling is not supported (see [glfw/issues/1168](https://github.com/glfw/glfw/issues/1168#issuecomment-1948239081)). 
I have 1.5 scaling, but all I can get from the api is 2.0.

Jacalz:
Unless @Geo25rey objects or beats me to it, I plan on trying to get this completed some time later this summer. Proper Wayland support, fractional scaling etc. are all things that will be very important for us in Fyne

EDIT: This may slip into August or September possibly. My summer is a bit busy and I have other things further ahead in the pipeline...

andydotxyz:
I'd like to do what I can to help get this over the line as we head into the new year. 3.4 is a big boost with many features and I think many will benefit.

Fyne wants to sponsor the projects we depend upon, however our sponsorship reduced a few months ago so this has not been possible.
However if a bounty will get this completed I will see what we can do.

What is the main blocker at this time?

Jacalz:
I ran out of time looking at this during summer and have since been extremely busy with studies and preparations for my master's thesis staring in January. As such, I will probably not be able to look at this until next summer.

As I see it, the remaining work is to make sure that Linux defaults to X11 + Wayland with optional `x11` and `wayland` build tags to only support one of them. Apart from that, it is just a case of making sure that it builds correctly on all operating systems and CPU architectures.

alexballas:
Hello, 

Here's my feedback

1. Updated the build tags in build.go so Linux defaults to both X11 and Wayland enabled. Users can still opt out with -tags x11 or -tags wayland
2. Fixed build constraints on native_linbsd_x11.go so X11 native Go functions are available when building with both X11 and Wayland enabled.
3. Changed SetIcon to use runtime platform detection via glfwGetPlatform() instead of a compile-time constant, so it works correctly when both backends are compiled in.
4. Removed a malformed  constraint line in not_wayland.go
5. Added the missing posix_module.c and posix_poll.c source files to the BSD cgo configuration.


And the diff:

```diff
diff --git a/v3.4/glfw/build.go b/v3.4/glfw/build.go
index 1327af4..4a48deb 100644
--- a/v3.4/glfw/build.go
+++ b/v3.4/glfw/build.go
@@ -28,9 +28,14 @@ package glfw
 // Linux Build Tags
 // ----------------
 // GLFW Options:
-#cgo linux,!wayland CFLAGS: -D_GLFW_X11
-#cgo linux,x11 CFLAGS: -D_GLFW_X11
-#cgo linux,wayland CFLAGS: -D_GLFW_WAYLAND
+// Default (no tags): both X11 and Wayland enabled
+// -tags x11: X11 only
+// -tags wayland: Wayland only
+// -tags x11,wayland: both enabled (explicit)
+#cgo linux,!x11,!wayland CFLAGS: -D_GLFW_X11 -D_GLFW_WAYLAND
+#cgo linux,x11,!wayland CFLAGS: -D_GLFW_X11
+#cgo linux,!x11,wayland CFLAGS: -D_GLFW_WAYLAND
+#cgo linux,x11,wayland CFLAGS: -D_GLFW_X11 -D_GLFW_WAYLAND

 // Linker Options:
 #cgo linux,!gles1,!gles2,!gles3,!vulkan LDFLAGS: -lGL
@@ -38,8 +43,11 @@ package glfw
 #cgo linux,gles2 LDFLAGS: -lGLESv2
 #cgo linux,gles3 LDFLAGS: -lGLESv3
 #cgo linux,vulkan LDFLAGS: -lvulkan
-#cgo linux,!wayland LDFLAGS: -lX11 -lXrandr -lXxf86vm -lXi -lXcursor -lm -lXinerama -ldl -lrt
+// X11 libraries (included when x11 tag is set or no display tags set)
+#cgo linux,!x11,!wayland LDFLAGS: -lX11 -lXrandr -lXxf86vm -lXi -lXcursor -lm -lXinerama -ldl -lrt
 #cgo linux,x11 LDFLAGS: -lX11 -lXrandr -lXxf86vm -lXi -lXcursor -lm -lXinerama -ldl -lrt
+// Wayland libraries (included when wayland tag is set or no display tags set)
+#cgo linux,!x11,!wayland LDFLAGS: -lwayland-client -lwayland-cursor -lwayland-egl -lxkbcommon -lm -ldl -lrt
 #cgo linux,wayland LDFLAGS: -lwayland-client -lwayland-cursor -lwayland-egl -lxkbcommon -lm -ldl -lrt

 // BSD Build Tags
diff --git a/v3.4/glfw/c_glfw_bsd.go b/v3.4/glfw/c_glfw_bsd.go
index 2c704f0..fe64608 100644
--- a/v3.4/glfw/c_glfw_bsd.go
+++ b/v3.4/glfw/c_glfw_bsd.go
@@ -22,6 +22,8 @@ package glfw
        #include "glfw/src/glx_context.c"
 #endif
 #include "glfw/src/null_joystick.c"
+#include "glfw/src/posix_module.c"
+#include "glfw/src/posix_poll.c"
 #include "glfw/src/posix_time.c"
 #include "glfw/src/posix_thread.c"
 #include "glfw/src/xkb_unicode.c"
diff --git a/v3.4/glfw/c_glfw_lin_wayland.go b/v3.4/glfw/c_glfw_lin_wayland.go
index 68e5ab1..4385cc3 100644
--- a/v3.4/glfw/c_glfw_lin_wayland.go
+++ b/v3.4/glfw/c_glfw_lin_wayland.go
@@ -1,5 +1,5 @@
-//go:build linux && wayland
-// +build linux,wayland
+//go:build (linux && !x11 && !wayland) || (linux && wayland)
+// +build linux,!x11,!wayland linux,wayland

 package glfw

diff --git a/v3.4/glfw/c_glfw_lin_x11.go b/v3.4/glfw/c_glfw_lin_x11.go
index e24465f..567b7ee 100644
--- a/v3.4/glfw/c_glfw_lin_x11.go
+++ b/v3.4/glfw/c_glfw_lin_x11.go
@@ -1,5 +1,5 @@
-//go:build (linux && !wayland) || (linux && x11)
-// +build linux,!wayland linux,x11
+//go:build (linux && !x11 && !wayland) || (linux && x11)
+// +build linux,!x11,!wayland linux,x11

 package glfw

diff --git a/v3.4/glfw/c_glfw_windows.go b/v3.4/glfw/c_glfw_windows.go
index 7ff980a..1928db5 100644
--- a/v3.4/glfw/c_glfw_windows.go
+++ b/v3.4/glfw/c_glfw_windows.go
@@ -7,6 +7,7 @@ package glfw
 #include "glfw/src/win32_time.c"
 #include "glfw/src/win32_thread.c"
 #include "glfw/src/win32_window.c"
+#include "glfw/src/win32_module.c"
 #include "glfw/src/wgl_context.c"
 #include "glfw/src/egl_context.c"
 */
diff --git a/v3.4/glfw/native_linbsd_x11.go b/v3.4/glfw/native_linbsd_x11.go
index 83125e2..192dc22 100644
--- a/v3.4/glfw/native_linbsd_x11.go
+++ b/v3.4/glfw/native_linbsd_x11.go
@@ -1,5 +1,5 @@
-//go:build (linux && !wayland) || (freebsd && !wayland) || (netbsd && !wayland) || (openbsd && !wayland)
-// +build linux,!wayland freebsd,!wayland netbsd,!wayland openbsd,!wayland
+//go:build (linux && !x11 && !wayland) || (linux && x11) || (freebsd && !wayland) || (netbsd && !wayland) || (openbsd && !wayland)
+// +build linux,!x11,!wayland linux,x11 freebsd,!wayland netbsd,!wayland openbsd,!wayland

 package glfw

diff --git a/v3.4/glfw/not_wayland.go b/v3.4/glfw/not_wayland.go
index b6b5de5..a46d02c 100644
--- a/v3.4/glfw/not_wayland.go
+++ b/v3.4/glfw/not_wayland.go
@@ -1,6 +1,5 @@
 //go:build !((linux && wayland) || (freebsd && wayland) || (netbsd && wayland) || (openbsd && wayland))
-// -build linux,wayland freebsd,wayland netbsd,wayland openbsd,wayland

 package glfw

-const WAYLAND = false
\ No newline at end of file
+const WAYLAND = false
diff --git a/v3.4/glfw/window.go b/v3.4/glfw/window.go
index 5c01082..92e7ecc 100644
--- a/v3.4/glfw/window.go
+++ b/v3.4/glfw/window.go
@@ -406,10 +406,14 @@ func (w *Window) SetTitle(title string) {
 // The desired image sizes varies depending on platform and system settings. The selected
 // images will be rescaled as needed. Good sizes include 16x16, 32x32 and 48x48.
 func (w *Window) SetIcon(images []image.Image) {
-       if (WAYLAND) {
+       platform := C.glfwGetPlatform()
+       if platform == C.GLFW_PLATFORM_WAYLAND {
                println("warning: Wayland: The platform does not support setting the window icon")
                return
        }
+       if platform == 0 {
+               return
+       }
        count := len(images)
        cimages := make([]C.GLFWimage, count)
        freePixels := make([]func(), count)
```

Jacalz:
That looks sensible. Thanks for continuing to work on this.

alexballas:
Hello @Geo25rey, will you be able to pick up the changes?

alexballas:
Hi everyone, I’ve opened PR #414 to pick up where we left off. I’ve addressed all the feedback and TODOs from this thread. Hopefully we can merge it soon!

PR #408, opened 2025-02-26
Replace Travis with GitHub Workflows and fix linter issues

Jacalz:
A small modernisation PR to make sure that we have some testing infrastructure up.


Jacalz:
All workflows are passing now :)

dmitshur:
Thanks for doing this.

I suggest reducing the scope to test only the module in v3.3 (and v3.4 when that’s in), not the top-level one with v3.2 and older. We’re not doing development for v3.2 and older, so it’s better not to modify them nor test them.

Jacalz:
Good point. Should be ready for a re-review now :)

.github/workflows/analysis.yml
```diff
@@ -0,0 +1,40 @@
...
+      run: rm -rf v3.0/ v3.1/ v3.2/
+
+    - name: Get dependencies
+      run: sudo apt-get update && sudo apt-get install gcc libgl1-mesa-dev libegl1-mesa-dev libgles2-mesa-dev libx11-dev xorg-dev
+
+    - name: Vet
+      run: go vet ./...
+
+    - name: Goimports
+      run: test -z "$(goimports -e -d . | tee /dev/stderr)"
```
dmitshur: I'm not sure how helpful it is to use goimports instead of using gofmt. I think gofmt would be helpful and sufficient.

But if you think there's enough upsides to goimports, let's keep it as is for now. Up to you.

.github/workflows/analysis.yml
```diff
@@ -0,0 +1,40 @@
...
+
+    - name: Gocyclo
+      run: gocyclo -over 15 .
```
dmitshur: Not sure how helpful this is. If you prefer to add it, I'm okay with it, but if it starts to produce reports that we deem to be false positives, let's drop it then.

.github/workflows/analysis.yml
```diff
@@ -0,0 +1,40 @@
...
+
+    - name: Staticcheck
+      run: staticcheck ./...
```
dmitshur: Sounds good to use staticcheck; it should have a very high signal to noise ratio, like vet. Given the slightly unusual nature of this library, if some of the checks cause friction, we can adjust them individually.

README.md
```diff
@@ -1,4 +1,8 @@
-# GLFW 3.3 for Go [![Build Status](https://travis-ci.org/go-gl/glfw.svg?branch=master)](https://travis-ci.org/go-gl/glfw) [![GoDoc](https://godoc.org/github.com/go-gl/glfw/v3.3/glfw?status.svg)](https://godoc.org/github.com/go-gl/glfw/v3.3/glfw)
+[![Tests](https://github.com/go-gl/glfw/actions/workflows/tests.yml/badge.svg)](https://github.com/go-gl/glfw/actions/workflows/tests.yml)
```
dmitshur, line 1: These links are broken now, but I understand they'll work after this is submitted and the actions run.

dmitshur (approved):
LGTM. Thanks.

README.md
Jacalz: Indeed. They should start working once this is merged

.github/workflows/analysis.yml
Jacalz: Agreed. I think it should work well for our use cases

.github/workflows/analysis.yml
Jacalz: I double the complexity budget so we aren't too strict. We have been using this in fyne since basiclaly forever and it is a good way of keeping functions short and readable

.github/workflows/analysis.yml
Jacalz: I reverted back to gofmt. It is probably fine as it is

Jacalz:
I'm having some issues with the go commands not working well across go modules. Will squash commits when it is all ready

PR #411, opened 2025-08-23
Fix for programs crassing when display goes idle

qbit:
As reported in https://github.com/fyne-io/fyne/issues/5899 , apps seem to crash when they are visible and the display goes through an idle -> wake change.

Running with this patch seems to have resolved my issue. This is a similar fix to what was done in kitty: https://github.com/kovidgoyal/kitty/commit/ab7f3310c4f7982026a897155eabb097683658b2


Jacalz:
I'd rather have this upstreamed into GFLW than have the fix added here. The Glfw sources in this repo are generated using a script and your changes will disappear when we regenerate as part of updating to a new release.

Jacalz:
Drew from the Fyne team has opened https://github.com/glfw/glfw/pull/2766 upstream so I'll close this per the comment above. Thanks for the contribution; it is much appreciated regardless!

Issue #188, opened 2017-04-05
Consider adding release tags


velovix:
With an official dependency management solution coming, I think it would be cool to see release tags on this repository. I'm not sure exactly how this should work, since this repository is already quasi-versioned with GLFW mainline. Perhaps what's here now could be tagged as v1.0.0 and the version number would only be bumped up when breaking changes or new features to existing bindings are introduced.

dmitshur:
> With an official dependency management solution coming, I think it would be cool to see release tags on this repository.

Can we wait until that solution comes out? As you said, the versioning for us is tricky, since it's tied to versions of GLFW upstream.

We don't yet know what requirements for the version numbers the dependency management solutions may require. Will it be semver? Will it be some variant thereof? How do the choices we make impact users? Etc.

I'd defer dealing with this until we have concrete reasons to do it, not speculative. As far as I know, there are no concrete benefits today, are there?

velovix:
I think the piece of mind that versioning gives users and the added flexibility it gives maintainers is a concrete reason in and of itself.

On the more theoretical side, my concern is that taking the "wait and see" attitude towards versioning in Go is self-defeating. Dep (the solution I'm referring to) needs to be vetted by the community first before being officially added to the Go toolchain, but the community is discouraged from vetting Dep by repositories with no release tags.

As for the issue of versioning syntax, Dep uses semantic versioning and I don't know of any reason why that would change.

dmitshur:
All right, I suppose I'm okay with doing it sooner then. I'm still not quite sold on the benefits, but if other people want this, I'm not against it.

We can use this issue to discuss and figure out what versioning system to use then.

GLFW itself seems to be using semver, but I'm not sure if we can use it directly, since there are some Go-specific changes in the API that may be independent of GLFW the C library's version. That said, using completely different version numbers may be confusing, especially since the major and minor version of GLFW are embedded in the import path of each `glfw` package.

One potential reference to look at is GopherJS, which is tied to Go minor versions. But the scheme GopherJS uses for its versions isn't semver. It's 1.x.y, where x is the Go minor version, and y is the GopherJS major version.

tapir:
We're already only breaking API with GLFW minor versions. So I'm not sure what's the advantage of having a separate version number as git tags. It would only cause confusion I think. My proposal is to use GLFW versions as the tags.

pwaller:
[...]

PR #414, opened 2026-02-26
Complete GLFW v3.4 integration and fix TODOs (supersedes #400)

### LINKED ISSUE #393: v3.4/glfw: add support for GLFW version 3.4 (and future 3.4.x versions)

alexballas:
This PR finalizes the migration to GLFW v3.4. It resolves the long-standing blockers from #400 and completes the remaining tasks required for a stable release.

This PR completes the GLFW v3.4 integration. It resolves the build failures and Wayland-related panics that stalled previous efforts.

Key Updates:
- Hybrid Linux Build: Both X11 and Wayland are now enabled by default with runtime switching. Use -tags x11 or -tags wayland to force a single backend.
- Build Fixes: Added missing C files (win32_module.c, posix_module.c, etc.) to fix linker errors on Windows, macOS, and BSD.
- Panic Prevention: Uses acceptError for FEATURE_UNAVAILABLE codes. This allows functions like SetIcon or SetOpacity to fail gracefully on platforms that don't support them (e.g., Wayland) instead of crashing.
- New Constants: Added all 3.4 Window Hints and Cursors (MousePassthrough, ResizeNWSE, etc.).
- Warning Fix: Wrapped _GNU_SOURCE in #ifndef to eliminate the "redefined" compiler warnings.

Thanks to @Geo25rey for the great start on this. I've finished the remaining work, including fixing the Windows/BSD linker errors and implementing the hybrid Linux builds. @Jacalz and @dmitshur, I also resolved the _GNU_SOURCE warnings and added the missing 3.4 hints/error handling to prevent Wayland panics. Ready for review!

@andydotxyz fyi

Fixes #393.
Closes #400.


alexballas:
Some more items:

- Added GLFW 3.4 platform API wrappers: GetPlatform, PlatformSupported, and Platform enum values.
- Added custom allocator support: Allocator + InitAllocator.
- Added Vulkan loader/presentation wrappers: InitVulkanLoader and GetPhysicalDevicePresentationSupport.
- Added direct error query API: GetError and NoError.
- Added missing GLFW 3.4 error mappings: CursorUnavailable, PlatformUnavailable (and handling in Init / standard cursor creation).
- Added missing window/input API pieces: Window.GetTitle, CursorCaptured, ContextNoError, AnyPosition, plus new 3.4 hints and hint values (ANGLE, Wayland, Win32, X11 Vulkan surface, etc.).
- Updated README to be in sync with v3.4 usage and changelog (including note that GLFW_UNLIMITED_MOUSE_BUTTONS is not in the vendored GLFW revision).

alexballas:
Brought this PR up to date with the latest go-gl commits. Everything should be in place now.

hackbar:
I tested this on my wayland system (KDE, debian 13.4) with opengl ES 3.2. It seems to work fine, but I noticed the window doesn't actually render until the first SwapBuffers call. With x11, it just shows a black window, but in wayland there's no actual window, just the app's icon in the taskbar. Not a huge issue, since I can just add a SwapBuffers call right after I make the window.

Also, I see the window isn't the size I asked for, but about 30% bigger in each dimension. Is this some hidpi thing? I am on a 4k 32" monitor

alexballas:
Hello @hackbar
Thanks for testing this :) 

What you describe matches typical Wayland behavior: unlike X11, a compositor may not show the toplevel until the client has attached and committed a real buffer, so seeing only the app icon until the first SwapBuffers is not surprising.

This is also reflected in the vendored GLFW 3.4 Wayland code. In `v3.4/glfw/glfw/src/egl_context.c:283`, there is even a note:
```
// NOTE: Swapping buffers on a hidden window on Wayland makes it visible
```

The size difference is most likely HiDPI / compositor scaling. In GLFW 3.4, window size and framebuffer size are separate: the window size is in screen coordinates, while the framebuffer size can be scaled for HiDPI.

Could you check what GLFW reports on your setup?

```
w, h := window.GetSize()
fbw, fbh := window.GetFramebufferSize()
xscale, yscale := window.GetContentScale()
fmt.Printf("window=%dx%d framebuffer=%dx%d scale=%.2fx%.2f\n", w, h, fbw, fbh, xscale, yscale)
```

If GetSize() matches the requested size, but GetFramebufferSize() is larger and GetContentScale() is above 1.0, then this is most likely HiDPI / compositor scaling on Wayland. If GetSize() itself is larger than requested, that would point to an actual sizing issue.

Thanks,
Alex

hackbar:
Hmm, it looks my GetSize and GetFramebuffer size match what I requested, and ContentScale is 1 in both dimensions.

This doesn't seem to match the window size I'm getting, which is 1.25 times larger. This does match the scale I set in kde's settings. Setting that to 100% does give me the same windows size as I requested.

So it seems like wayland is doing some kind of scaling for me automatically? Is this a thing, or is there any way to change the setting?

alexballas:
Could you check one more thing on top of the previous printout: what does `glfw.GetPlatform()` report after `Init()`?
It would also help to print the same values once more after the window is actually mapped, i.e. after one initial `SwapBuffers()` + `PollEvents()`.

If native Wayland still reports `window == framebuffer` and `scale == 1.0` even after the first frame, while KDE at 125% still makes the window appear larger, then this points to upstream GLFW / compositor scaling behavior rather than `go-gl` itself.

hackbar:
GetPlatforms return 0x60003.

Even after the first buffer swap it's showing the framebuffer and window are the same size (what I requested) and the scale is 1.0.

I'll try to see how KDE scaling works. I'd imagine the kwin compositor does things the standard wayland way, but maybe it's doing something weird.

alexballas:
Could you check whether the compositor is advertising the needed protocols?
`wayland-info | grep -E 'wp_fractional_scale_manager_v1|wp_viewporter'`
and one step deeper
`WAYLAND_DEBUG=1 ./app 2>&1 | grep -E 'wp_fractional_scale|preferred_scale|wp_viewporter'`

hackbar:
Yeah, it has both those extensions:
interface: 'wp_viewporter',                              version:  1, name:
 6
interface: 'wp_fractional_scale_manager_v1',             version:  1, name:
 8

alexballas:
A good next check is whether KWin is actually sending the fractional-scale callback for the window (the second grep I sent). GLFW should request it, but GetContentScale() will stay at 1.0 unless the compositor replies with preferred_scale. If you run the test with WAYLAND_DEBUG=1, we can check whether GLFW sends get_fractional_scale and whether KWin sends preferred_scale back.

hackbar:
Which tests? just printing out the window/framebuffer/scaling sizes?

hackbar:
here's the logs for printing the sizes etc: https://pastebin.com/zP2gYkZt

hackbar:
Is there anything else I can do to help test this? Other than than the scaling, which I don't really mind once I figured out what was happening (my app is primarily fullscreen), it's working well for me.

Jacalz:
I have been meaning to review this but I'm afraid that I have very little time for open source at the moment due to writing my Master's thesis.

rcoreilly:
FWIW, I have been using this PR in https://cogentcore.org apps for a week or so without any issues on my macbook. Given how difficult it is to actually use the PR code from another package (because 3.4 doesn't exist in any package that go modules know about), I haven't tested on other platforms. And also given that using this requires changing the import paths, perhaps it makes sense at this point to fast-track the approval, and then plan to address any issues that come up later?

alexballas:
Agree. Since this requires explicit import changes, it won’t break existing integrations by default. We can address edge cases as they arise.

hackbar:
Yeah, I think that makes a lot of sense.

Jacalz:
Perhaps @dmitshur has more time to look at it?

README.md
```diff
@@ -137,7 +174,7 @@ func main() {
 
 
 ### GLFW 3.2 Specfic Changes
-- Easy `go get` installation. GLFW source code is now included in-repo and compiled in so you don't have to build GLFW on your own and distribute shared libraries. The revision of GLFW C library used is listed in [GLFW_C_REVISION.txt](https://github.com/go-gl/glfw/blob/master/v3.3/glfw/GLFW_C_REVISION.txt) file.
+- Easy `go get` installation. GLFW source code is now included in-repo and compiled in so you don't have to build GLFW on your own and distribute shared libraries. The revision of GLFW C library used is listed in [GLFW_C_REVISION.txt](https://github.com/go-gl/glfw/blob/master/v3.4/glfw/GLFW_C_REVISION.txt) file.
```
dmitshur: Since this is in the 3.2 section, pointing to the v3.4/glfw/GLFW_C_REVISION.txt file doesn't seem right. Its previous value of v3.3 was also not right. I think the right target is https://github.com/go-gl/glfw/blob/master/v3.2/glfw/GLFW_C_REVISION.txt.

scripts/generate-buildable-source-code.sh
```diff
@@ -0,0 +1,118 @@
...
+
+generate_dummy_go_files() {
+    local glfw_root="$1"
+    local deps_import_root="$2"
+
+    find "$glfw_root/deps" "$glfw_root/include" "$glfw_root/src" -type d -print0 |
+        while IFS= read -r -d $'\0' d; do
+            cat > "$d"/dummy.go <<'EOF'
+//go:build required
+// +build required
```
dmitshur, line 36: Since the go directive is 1.19+ everywhere by now, we can start dropping the old +build lines in favor of the new go:build ones. Just noting this, but no need to do that in this PR, we can do it in a future one.

v3.4/glfw/vulkan.go
```diff
@@ -0,0 +1,132 @@
...
+	if deviceValue.Kind() != reflect.Ptr {
+		return false, fmt.Errorf("vulkan: device is not a VkPhysicalDevice (expected kind Ptr, got %s)", deviceValue.Kind())
+	}
+
+	supported = glfwbool(C.glfwGetPhysicalDevicePresentationSupport(
+		(C.VkInstance)(unsafe.Pointer(instanceValue.Pointer())),
+		(C.VkPhysicalDevice)(unsafe.Pointer(deviceValue.Pointer())),
+		C.uint32_t(queueFamily),
+	))
+	if acceptErr := acceptError(APIUnavailable); acceptErr != nil {
```
dmitshur: Minor nitpick, it should be simpler and consistent with similar code that uses `acceptError` function elsewhere to rename `acceptErr` variable to just `err` here.

dmitshur (approved):
Thanks very much @alexballas for getting this PR prepared and addressing remaining comments that were left on PR #400. Thank you @Geo25rey for getting PR #400 very far.

I've reviewed and been testing the head of PR #400 locally for a long while, and I haven't spotted problems. I reviewed the diff of this PR against what I already reviewed earlier, and the changes look okay to me. I'm also not seeing problems from brief testing on darwin/arm64 locally. Thanks to others who've also looked and tested this PR.

I left a few minor review comments, but they're very minor. I think it's fine for them to be addressed afterwards, so I'll go ahead and merge this now.

I also plan to tag the first pre-release version of the new v3.4/glfw module as `v0.1.0-pre.1` per plan in https://github.com/go-gl/glfw/issues/188#issuecomment-2121640512 soon, perhaps tomorrow. **Edit:** [Done](https://pkg.go.dev/github.com/go-gl/glfw/v3.4/glfw?tab=versions). We'll do further tags, potenitally more pre-release tags and eventually `v0.1.0`.

scripts/generate-buildable-source-code.sh
dmitshur: Marking as resolved; this note isn't meant to be in scope of this PR.

README.md
dmitshur: Done in commit fb51f1547102b67bbc96e40e0ed733a125c2dc49.

v3.4/glfw/vulkan.go
dmitshur: Done in commit cc44bab941ad35dfd2c2c7e34ba6ffb21519d7e7.

hackbar:
This is still working great - thanks!

I'll file a bug to track the KDE scaling issue. I meant to do that, but forgot.

PR #415, opened 2026-04-06
README.md: remove -u flag from go get

dmitshur:
The -u flag was needed back in GOPATH mode to reliably fetch the latest version of the target package and its dependencies. There's no need for it in module mode, since the default [version query](https://go.dev/ref/mod#version-queries) is `@upgrade` (which acts a lot like `@latest`). Leave it out.


Jacalz (approved):

dmitshur:
Thanks.

PR #417, opened 2026-04-10
Fix 32bit build

### LINKED ISSUE #416: v3.4 GLFW_ANY_POSITION on arm 32bit
I'm getting the following error:

```
# github.com/go-gl/glfw/v3.4/glfw
Error: ../../../go/pkg/mod/github.com/go-gl/glfw/v3.4/glfw@v0.1.0-pre.1/window.go:208:20: cannot use (_Ciconst_GLFW_ANY_POSITION) (untyped int constant 2147483648) as int value in constant declaration (overflows)
Error: Process completed with exit code 1.
```

The binding does:

```
AnyPosition int = C.GLFW_ANY_POSITION
```

And upstream defines:

```
#define GLFW_ANY_POSITION 0x80000000
```

So I guess it overflows the int size on 32bit?

See: https://github.com/libretro/ludo/actions/runs/24262354910/job/70849082030#step:20:283

kivutar:
Fixes https://github.com/go-gl/glfw/issues/416

### DIFF
--- a/v3.4/glfw/window.go
+++ b/v3.4/glfw/window.go
@@ -205,7 +205,7 @@ const (
 
 	// AnyPosition is a hint value for PositionX and PositionY that lets the
 	// window manager decide the initial window position.
-	AnyPosition int = C.GLFW_ANY_POSITION
+	AnyPosition int = -1 << 31 // C.GLFW_ANY_POSITION
 )
 
 // Window represents a window.


Jacalz (commented):
Thanks. The error is gone but the fact that 0x800... overflows an int32 still remains as far as I can see. Would it not (I haven't checked the code) be possible to change the type on our end to int64 instead? Seems like possibly a better fix to me as that constant still would be 0 on 32 bit systems otherwise. Do they place windows in the wrong spot then?

v3.4/glfw/window.go
```diff
@@ -205,7 +205,7 @@ const (
 
 	// AnyPosition is a hint value for PositionX and PositionY that lets the
 	// window manager decide the initial window position.
-	AnyPosition int = C.GLFW_ANY_POSITION
+	AnyPosition int = -1 << 31 // C.GLFW_ANY_POSITION
```
Jacalz, line 208: If the idea above is unfeasable, maybe it's possible to force it to convert to int with some kind of cast so we don't need to change it manually going forward?

kivutar: I was also thinking about int64, but it's hard because it would change the public API of WindowHint.

-1 << 31 is not overflowing though, it's value is -2147483648, a signed int32 with the same 32-bit bit pattern as 0x80000000.

We can't cast 0x80000000 to this, because int32(0x80000000) would overflow.

The bit pattern it 10000000 00000000 00000000 00000000
If you see it as a uint32, it's 0x80000000 or 2147483648
If you see it as a int32, it's -2147483648

Go complains about it, unlike C, because it sees 0x80000000 as a positive constant. C is not that strict.
-1 << 31 or -2147483648 is a negative constant, so Go will accept it in an int.

I don't know a way to express that the want 10000000 00000000 00000000 00000000 from C.GLFW_ANY_POSITION. For vars maybe, but for const, I'm not aware of a way.

alexballas: Our options are fairly limited here, and this PR feels like a reasonable compromise.

andydotxyz:
Having upgraded to the 3.4 tree for Fyne v2.8.0 we are unfortunately blocked by this PR.
@Jacalz what can we do to get this closed out? Chatting with Alex it seems like the PR as-is should be kept - it works and it un-breaks compilation... If a better solution is sought could it be explored after the compile is working again?

v3.4/glfw/window.go
Jacalz: Alright, good points. Let's merge and get 32 bits working again!

Jacalz (approved):

Issue #419, opened 2026-05-10
v3.4/glfw: Linux, no buildtags does not expose `GetWaylandDisplay`


oliverbestmann:
While both PlatformWayland and PlatformX11 are reported as supported, I can only get a `C.Display` for X11. The build tags for `native_linbsd_x11.go` are set to `//go:build (linux && !x11 && !wayland)` while for the wayland file they are set to `//go:build (linux && wayland)` .. Is this on purpose or by accident? 

So, if I want to support both platforms I have to build with `x11,wayland`, is that correct? I would have expected the default build to support both platform and to be restrict to a single platform by using the build tags.

Edit: Actually, it auto selects wayland, but due to the constraints I can not get the handle for the surface. I'll create a PR for this.

dmitshur:
I understand PR #420 fixed this, so closing.

PR #420, opened 2026-05-10
glfw/v3.4: Improve build constraints for x11 & wayland handling

oliverbestmann:
Fixed the build tags for linux to correctly expose surfaces for x11 and wayland.

The logic is now:

* No build tags: expose both x11 and wayland
* Either `x11` or `wayland`: expose only one
* Both `x11,wayland`: expose both x11 and wayland

I've removed the old build constraint syntax as the minimum go version is defined as 1.19 anyways.

### DIFF
--- a/v3.4/glfw/native_linbsd_wayland.go
+++ b/v3.4/glfw/native_linbsd_wayland.go
@@ -1,5 +1,4 @@
-//go:build (linux && wayland) || (freebsd && wayland) || (netbsd && wayland) || (openbsd && wayland)
-// +build linux,wayland freebsd,wayland netbsd,wayland openbsd,wayland
+//go:build (linux || freebsd || netbsd || openbsd) && ((!x11 && !wayland) || wayland)
 
 package glfw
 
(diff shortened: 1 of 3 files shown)


rcoreilly:
Fixes #419

This appears to be consistent with the intent of the original 3.4 PR (#414)

> Hybrid Linux Build: Both X11 and Wayland are now enabled by default with runtime switching. Use -tags x11 or -tags wayland to force a single backend.

@dmitshur would it be possible to give this a quick look? It is preventing effective use of the webgpu library https://github.com/oliverbestmann/webgpu due to constraints on the use of replace directives in go.mod..

andydotxyz:
I've tested that on Fyne's develop and this fixes the issue and would be highly desirable!

For now Fyne apps in auto-detect cannot use Wayland windows as I pass a 0 surface ID instead...

Jacalz:
Perfect. Thank you!

Jacalz (approved):

PR #423, opened 2026-07-22
glfw/v3.4: fix preserving wayland headers in vendor

andydotxyz:
Does what it says on the tin :)


andydotxyz:
Just a little nudge here. Anyone not incorporating a workaround like Fyne has done will get a compile error if they vendor...

dmitshur (approved):
Thanks.

